PR #40, opened 2021-11-03
Add query_id to response data for run_result

### LINKED ISSUE #7: Include the snowflake query id in the run results output
### Describe the feature
As an analytics engineer I want a quick way to get the query id of a model run (or even test/seed/snapshot) so I can quickly see the query in the snowflake UI

We've implemented a macro that runs in the post-hook that will get the full url to the query and log it to console. This works great except that we have to call `SELECT last_query_id()` on every model run and more importantly **the hook does not execute if the query fails**

I acknowledge that the compiled SQL is available as an alternative but we run airflow on kubernetes so the artifacts are not readily available. We do persist the artifacts in s3 but I think it would be an improved experience to go from airflow log to snowflake ui in one click.

### Additional context
I can see that the snowflake query id is being logged for failed queries if you enable debug mode but the output is far too verbose for this use case. I would want just the query url/id logged on every run so that a user can go from log => snowflake ui with the fewest steps possible.

### Who will this benefit?
I think this would benefit all dbt-snowflake users that use the snowflake ui for debugging/performance monitoring.

I wouldn't mind contributing but I would need some guidance on what option is most amenable to dbt's design. I'm not sure if the plugins are meant to interact with the run results functionality in dbt core.

joshuataylor:
resolves #7

### Description

I'm building an application to parse `run_result.json`, and being able to get the `query_id` out would be amazing. I noticed that there is an issue that had the exact same idea.

This now adds query_id to `run_result.json` as follows:
```
      {
         "status":"success",
         "timing":[
            {
               "name":"compile",
               "started_at":"2021-11-03T11:01:00.755167Z",
               "completed_at":"2021-11-03T11:01:00.762187Z"
            },
            {
               "name":"execute",
               "started_at":"2021-11-03T11:01:00.763095Z",
               "completed_at":"2021-11-03T11:01:03.140846Z"
            }
         ],
         "thread_id":"Thread-3",
         "execution_time":3.4953978061676025,
         "adapter_response":{
            "_message":"SUCCESS 1",
            "code":"SUCCESS",
            "rows_affected":1,
            "query_id":"01a00a55-0401-4dec-0001-1c830efa5e8a"
         },
         "message":"SUCCESS 1",
         "failures":null,
         "unique_id":"model.my_new_project.stg_customers"
      },
```

As noted in the issue above, we need to add tests to this probably as well.

### Checklist


jtcohen6 (commented):
@joshuataylor Thanks for the contribution, this is looking really solid!

I think an integration test is a good idea, and should be really simple to set up. Something like:
```python
@use_profile('snowflake')
    def test_snowflake_adapter_methods(self):
        results = self.run_dbt()
        assert results[0].adapter_response['query_id'] is not None
```
Let me know if you have questions, I can try to give more pointers.

anthu:
Hey @joshuataylor ,

Happy New Year and I hope you're doing good.
Also from my side, thanks a lot for this super helpful addition to the adapter!

What do you think of the proposed test above? 
In case you're just busy and don't have capacity to finish this off, I'm offering you to add this test on my own so we can ship this feature with the next release. 

Just let me know.

joshuataylor:
Ah sorry!

I totally missed this, I'll get this done within the next 24 hours or so.

anthu:
Hey @jtcohen6 ,

Sorry for pinging you directly. What do you think of the test contributed?
Maybe we could squash in this enhancement for the next release?

Best,
Anton

dbt/adapters/snowflake/connections.py
```diff
@@ -28,6 +28,10 @@
 _TOKEN_REQUEST_URL = 'https://{}.snowflakecomputing.com/oauth/token-request'
 
 
+@dataclass
+class SnowflakeAdapterResponse(AdapterResponse):
+    query_id: Optional[str] = None
```
VersusFacit, line 33: "Oh pedantry of types! you are my curse." ~~ Shakespeare or something

Concerning [the definition of the `AdapterResponse` class](https://github.com/dbt-labs/dbt-core/blob/434f3d26789d4bed3f0057c7c7d898e9dbf6687f/core/dbt/contracts/connection.py#L30), notice the `_msg` field is just a string and not an `Optional`. Why? Because there should always be a message string. And indeed, there's one even for bugged queries:

image

Can a query_id be null? I'm expecting the answer to be "no" ([stored procedures](https://community.snowflake.com/s/article/Why-does-Snowflake-LAST-QUERY-ID-function-returns-NULL-when-I-try-to-get-the-query-ID-which-is-executed-in-stored-procedure) notwithstanding). If so, in the spirit of being good abstracters, let's remove the Optional :) Per this [stackoverflow thread](https://stackoverflow.com/questions/63892939/is-query-id-unique-among-all-the-snowflake-accounts?rq=1), looks like this field is even UUID-unique across all accounts -- cool!

VersusFacit (commented):
I'll activate the testing workflow on this. Do please ping me directly. Let's get this wrapped up and merged 🚀

dbt/adapters/snowflake/connections.py
VersusFacit: If we change the type here, change the test predicate to a plain `assert` statement instead of having the `is not` predicate (i.e. make sure a query_id is actually present!).

VersusFacit:
Thanks for all the effort here. I merged this over in #109 with credit given to Josh as the primary author of the changes. :D

DemchenV:
Hi all! I'm using dbt-snowflake version 1.1.0 and want to get query_id for tests, but run_result.json contains only query_id for runs. As I understand, query_id for tests was also added in 1.1.0, am I right?

PR #46, opened 2021-11-08
Fix tests failing due to dbt-core changes

jtcohen6:


jtcohen6:
This previously failed because of an accidental change to `get_batch_size()` in `dbt-core`, which removed the `return() wrapping (i.e. the fix in https://github.com/dbt-labs/dbt-core/pull/3942). We just merged https://github.com/dbt-labs/dbt-core/pull/4240, so I'm rerunning the test in the hopes it works now :)

emmyoop (approved):
LGTM

PR #74, opened 2021-12-21
Use case insensitive comparison

### LINKED ISSUE #73: Use lowercase / case-insensitive comparison when persisting column-level comments for views
See https://github.com/dbt-labs/dbt-snowflake/issues/72 for context

https://github.com/dbt-labs/dbt-snowflake/blob/e75986f0b81749bc790f52dcea4ecd1c1a7f3f2a/dbt/include/snowflake/macros/adapters.sql#L47-L50

This check should use `|lower` instead of `|upper`, since Snowflake columns are almost always uppercase.

```python
model_columns:  {'id': {'name': 'id', 'description': 'this is a cool column', 'meta': {}, 'tags': []}}
query_columns:  ['ID']
```

That will enable users to persist descriptions for columns defined in lowercase:
```yml
version: 2

models:
  - name: my_cool_view_model
    description: this is a cool model
    config:
      materialized: view
      persist_docs:
        relation: true
        columns: true
    columns:
      - name: id # instead of ID
        description: this is a cool column
```

anthu:
resolves #73

### Description

Using case insensitive comparison, provided by Jinja

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -45,7 +45,7 @@
 {% macro get_persist_docs_column_list(model_columns, query_columns) %}
 (
   {% for column_name in query_columns %}
-    {% if (column_name|upper in model_columns) or (column_name in model_columns) %}
+    {% if (column_name|upper in model_columns|upper) %}
       {{ get_column_comment_sql(column_name, model_columns) }}
     {% else %}
       {{column_name}}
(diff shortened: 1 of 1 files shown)


anthu:
== Hold pls - looking into the the test failures

VersusFacit:
Hey @anthu , thanks for jumping on this one. Getting caught up with the context here.

Did you happen to see the comment tacked on by a community member, namely [this one](https://github.com/dbt-labs/dbt-snowflake/issues/73#issuecomment-1004824012)? I myself am a big fan of the maxim it's good to be cautious. 

In particular, your logic is more concise, but it leads to undefined outcomes in some "out there" edge cases. Check out the id example in the issue linked just prior. It all makes things a bit less readable for intent (unless you want to comment the intent -- either way works to me).

The one that worries me is the possibility that ID and id could be present in the _same_ table. That leads to some very strange behavior.

Wanna' take a swing at bulletproofing these out? 😄

jtcohen6:
I believe this was resolved by #161 - going to close!

PR #93, opened 2022-02-09
Remove use of temp table for Snowflake incremental merge

### LINKED ISSUE #92: [CT-185] Remove use of temporary table for merge incremental strategy in Snowflake
### Describe the feature
The MERGE command in Snowflake supports the use of arbitrary SELECT queries in the USING clause. If we were to adopt this approach for the merge incremental strategy, it remove a database write step that we are currently performing due to the creation of a temporary table to store the data to-be-merged.

The alternative is leaving the status quo. This would represent a performance improvement for the adapter when using the merge strategy.

### Additional context
Please include any other relevant context here.

### Who will this benefit?
Any customer who is working with incremental tables that have very large incremental loads stand to see a performance improvement with this.

Yes, to an extent. I can get some of the way, but would probably need assistance getting this to a merge-able state.

jnatkins:
resolves #92 
[CT-185] Remove use of temporary table for merge incremental strategy in Snowflake

### Description

This is a bit work-in-progress, but wanted to put something in place to start discussion around what the right resolution would look like. Apologies for whitespace changes...my editor apparently likes to clear out whitespace on empty lines.

### Checklist

- [] This PR includes tests, or tests are not required/relevant for this PR

[CT-185]: https://dbtlabs.atlassian.net/browse/CT-185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -24,8 +23,18 @@
   {% endif %}
 {% endmacro %}
 
+{% macro dbt_snowflake_create_temp_relation(strategy, tmp_relation, sql) %}
```
jtcohen6, line 26: I spent a few minutes thinking about whether it makes sense to combine `dbt_snowflake_get_incremental_sql` + `dbt_snowflake_create_temp_relation`, to avoid writing the same condition check for `strategy` twice.

In order to make that happen, we'd need to wrap these materialization steps into a third macro that gets called by both. Something like:
```sql
{% macro prep_for_incremental_sql(on_schema_change, tmp_relation, target_relation) %}
    {% do adapter.expand_target_column_types(
           from_relation=tmp_relation,
           to_relation=target_relation) %}
    {#-- Process schema changes. Returns dict of changes if successful. Use source columns for upserting/merging --#}
    {% set dest_columns = process_schema_changes(on_schema_change, tmp_relation, target_relation) %}
    {% if not dest_columns %}
      {% set dest_columns = adapter.get_columns_in_relation(target_relation) %}
    {% endif %}
    {{ return(dest_columns) }}
{% endmacro %}
```
Then we have a macro that does:
```sql
{% macro dbt_snowflake_get_incremental_sql(strategy, tmp_relation, target_relation, unique_key, dest_columns) %}
  {% if strategy == 'merge' %}
    {% do run_query(create_view_as(tmp_relation, sql)) %}
    {% set dest_columns = prep_for_incremental_sql(on_schema_change, tmp_relation, existing_relation) %}
    {% do return(get_merge_sql(target_relation, tmp_relation, unique_key, dest_columns)) %}
  {% elif strategy == 'delete+insert' %}
    {% do run_query(create_table_as(True, tmp_relation, sql)) %}
    {% set dest_columns = prep_for_incremental_sql(on_schema_change, tmp_relation, existing_relation) %}
    {% do return(get_delete_insert_merge_sql(target_relation, tmp_relation, unique_key, dest_columns)) %}
  {% else %}
    {% do exceptions.raise_compiler_error('invalid strategy: ' ~ strategy) %}
  {% endif %}
{% endmacro %}
```
And finally, in the materialization itself:
```sql
{% else %}
  {% set build_sql = dbt_snowflake_get_incremental_sql(strategy, tmp_relation, target_relation, unique_key, dest_columns) %}
{% endif %}
```
That might be totally overkill! Or it might leave the materialization a bit leaner + more modular

jtcohen6 (commented):
Nice work on an elegant solution! This achieves the desired results with minimal code changes. Have you been able to confirm that swapping the temp table for a view achieves the hoped-for performance boost? 

My original instinct here was to avoid creating an intermediate relation at all (whether temp table or view), if the `merge` strategy is being used. To that end, we'd need to determine if we can use `get_columns_in_query` to power `on_schema_change` (if set) and `expand_target_column_types` (always), same as we do for persisting column-level comments on views:

https://github.com/dbt-labs/dbt-snowflake/blob/9534b54760989e9d084e06d163fc9469481c1729/dbt/include/snowflake/macros/adapters.sql#L69

In both cases, we'd be "running" the model SQL twice—the first time either with `where false limit 0`, or saving the query in a view—the second time during the actual `merge`. Creating the view does seem like the less janky of the two potential approaches.

@VersusFacit Let's grab some time to talk more about this :) I'm curious to hear what you think of the proposed structure.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -24,8 +23,18 @@
...
+  {% if strategy == 'merge' %}
+    {% do return(create_view_as(tmp_relation, sql)) %}
+  {% elif strategy == 'delete+insert' %}
+    {% do return(create_table_as(True, tmp_relation, sql)) %}
```
niall-wm, line 30:
```suggestion
    {% if unique_key %}
      {% do return(create_table_as(True, tmp_relation, sql)) %}
    {% else %}
      {# If there is not a unique key, then there is only an insert step, so materialize as view #}
      {% do return(create_view_as(tmp_relation, sql)) %}
    {% endif %}
```

dbt/include/snowflake/macros/materializations/incremental.sql
niall-wm, line 26:
```suggestion
{% macro dbt_snowflake_create_temp_relation(strategy, unique_key, tmp_relation, sql) %}
```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -43,18 +52,18 @@
...
     {% set build_sql = create_table_as(False, target_relation, sql) %}
-  
+
   {% elif full_refresh_mode %}
     {% set build_sql = create_table_as(False, target_relation, sql) %}
-  
+
   {% else %}
-    {% do run_query(create_table_as(True, tmp_relation, sql)) %}
+    {% do run_query(dbt_snowflake_create_temp_relation(strategy, tmp_relation, sql)) %}
```
niall-wm, line 66:
```suggestion
    {% do run_query(dbt_snowflake_create_temp_relation(strategy, unique_key, tmp_relation, sql)) %}
```

niall-wm:
I've just tested this out on a model which uses the delete+insert strategy without a unique key set (so it actually only inserts) and it cut down the run time from 106s to 66s. I've made a couple of suggested changes which allow for a view to be made in a delete+insert strategy provided that there are no unique_keys.

dbt/include/snowflake/macros/materializations/incremental.sql
NiallRees, line 26:
```suggestion
{% macro dbt_snowflake_create_temp_relation(strategy, unique_key, tmp_relation, sql) %}
```

dbt/include/snowflake/macros/materializations/incremental.sql
NiallRees, line 66:
```suggestion
    {% do run_query(dbt_snowflake_create_temp_relation(strategy, unique_key, tmp_relation, sql)) %}
```

dbt/include/snowflake/macros/materializations/incremental.sql
NiallRees, line 30:
```suggestion
    {% if unique_key %}
      {% do return(create_table_as(True, tmp_relation, sql)) %}
    {% else %}
      {# If there is not a unique key, then there is only an insert step, so materialize as view #}
      {% do return(create_view_as(tmp_relation, sql)) %}
    {% endif %}
```

NiallRees:
I've just tested this out on a model which uses the delete+insert strategy without a unique key set (so it actually only inserts) and it cut down the run time from 106s to 66s. I've made a couple of suggested changes which allow for a view to be made in a delete+insert strategy provided that there are no unique_keys. Great work

McKnight-42 (approved):
The logic here is very easy to follow, some sound checks, tested locally 👍

nathaniel-may (commented):
Leaving my approval post-merge since I did review this in a live sync session earlier.

PR #96, opened 2022-02-17
Test incremental model with unique key both as string and list

### LINKED ISSUE #91: [CT-129] Add integration test for new list input for unique ID
This PR (dbt-labs/dbt-core#4618) added the ability to pass a list for unique ID. We would like to have an integration test for Snowflake that verifies that holds true here

VersusFacit:
Resolves #91 
Adds tests to this adapter for unique key changes in [PR 4618](https://github.com/dbt-labs/dbt-core/pull/4618) in dbt-core.

### Description

Adds three classes to mirror the branching paths (unique_key as string, list, both). I tried not to duplicate tests already found in dbt-core. Where possible, I followed existing conventions (e.g. using expected SQL models for comparison). There's a ton of side effects by virtue of model building, but I've attempted to keep things readable by separating out steps into functions.

To help, I've added intention based documentation where possible, but feel free to request more!

I've written a set of tests that not only covers several ways to specify `unique_key` but also simulates real-world incremental update behavior. In doing so, these tests ensure model build and uniqueness _semantics_ are carried out. I mocked up examples in a jaffle_shop first since incremental behavior is a bit wonky at first.

Because incremental model behavior requires multiple runs to display, I opted for actual model files as opposed to queries in code.

edit: Sorry for the force pushes. I had to tease out last minute an error due to mission `unique_key` in an integration test. It was passing locally but here until I got the parameters _just right_. Just one of those bugs! Should be reliably resolved though. And then this PR fell prey to the Jinja dependency failure. Integration tests are passing locally.

Ps: I have also been requested to create documentation for Leona to update community docs and am midway through its production. I've already met with @runleonarun  synchronously and showed her a lot of the new featureset added. Would be happy to share this with the team.

### Checklist


VersusFacit:
Currently suffering from the MarkupSafe==2.0.1, but integration tests are running smoothly locally.

kwigley (commented):
This looks good to me! I appreciate the edge cases covered here. No request changes from me, but I'm leaving approval up to another reviewer 👍

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -0,0 +1,270 @@
...
+            relation=incremental_model
+        )
+
+        self.run_test(expected_fields, test_case_fields)
+
+    @use_profile('snowflake')
+    def test__snowflake_bad_unique_key(self):
+        '''using a unique key not in seed or derived CTEs leads to an error'''
+        with self.assertRaises(AssertionError) as exc:
+            self.run_dbt(['run', '--select', 'not_found_unique_key'])
```
kwigley, line 165: One suggestion:

It might be nice to assert a more meaningful error is raised here. You can pass `expect_pass=False` to this method, which should prevent the test harness from raising an `AssertionError`. I'm not sure exactly what error to expect, though.

VersusFacit: Awesome! I didn't know about this but would love to check for something more specific than the generic assertion error. I'll get into the debugger and find out the exception thrown for both of these fail tests.

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -0,0 +1,270 @@
...
+class TestIncrementalUniqueKey(DBTIntegrationTest):
+    @property
+    def schema(self):
+        return 'incremental_unique_key'
+
+    @property
+    def models(self):
+        return 'models'
+
+    def count_test_relations(self, seed, incremental_model):
```
ChenyuLInx, line 23: Should we name this to something like create seed or create incremental models? count doesn't cover everything this function does. I feel like we can also just put this in the setup with some comment since this function is only used in one place

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -0,0 +1,270 @@
...
+        '''Idempotently create some number of seeds and incremental models'''
+        seed_result_set = self.run_dbt(
+            ['seed', '--select', seed, '--full-refresh']
+        )
+        model_result_set = self.run_dbt(
+            ['run', '--select', incremental_model, '--full-refresh']
+        )
+        return len(seed_result_set), len(model_result_set)
+
+    def get_seed_rows_after_update(self, seed, update_sql_file):
```
ChenyuLInx, line 33: Same with this one, it doesn't have to be it's own function. But more of a nitpicking here

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -0,0 +1,270 @@
...
+        seed_rows = self.get_seed_rows_after_update(
+            seed=seed, update_sql_file=update_sql_file
+        )
+        inc_test_model_count = self.update_incremental_model(
+            incremental_model=incremental_model
+        )
+
+        return (seed_count, model_count, seed_rows, inc_test_model_count)
+
+    def run_test(self, expected_fields, test_case_fields):
```
ChenyuLInx, line 62: This part doesn't run things but more like comparing result? Should we consider updating the name a bit?

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -0,0 +1,270 @@
...
+            inc_test_model_count=1, opt_model_count=opt_model_count,
+            relation=relation
+        )
+
+class TestIncrementalWithoutUniqueKey(TestIncrementalUniqueKey):
+    @use_profile('snowflake')
+    def test__snowflake_no_unique_keys(self):
+        '''with no unique keys, seed and model should match'''
+        seed='seed'
+        seed_rows=8
```
ChenyuLInx, line 108: I saw that most seed rows are either 7 or 8, do you think it would make sense to modify the test case so that we have more diverse number for different situation?

ChenyuLInx (commented):
This looks great. It covered all of the possible cases as far as I can tell. Added a few nitpicking comment for naming stuff.

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
VersusFacit: Hey Chenyu, I went back and forth on this myself. Here's my thought.

In the abstraction of the code itself, the primary thing occurring is obtaining the number of models built (a number returned is a sign we are successful in building our code). The state of the models on the database itself is actually a "side effect" in my opinion. The returned model count is a number only obtained because I was able to successfully create the database. Hence, the function should be named in a manner that reflects the count of test relations (seeds and tables) built.

Do you buy that? Got a better name in mind? Contrary thoughts? :)

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
VersusFacit: Same remark about focusing on the in-code effect of knowing what the updates are doing rather than focusing on the updates themselves which are database side effects.

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
VersusFacit: That's a fair callout! Hmmm, maybe something like test_scenario_correctness?

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
VersusFacit: I could. It would just mean adding another seed file. Not sure if that's needed now. Really, I set this up now so that we could easily add things in the future. See, this incremental test is mostly a matter of the combination between unique keys and either a simple insert/a merge. And so, my thinking went, "We only need two test scenarios; the model configs are where a lot of the testing will take place."

That make your concerns go away or nah, you still think we should diversify out the underlying test scenarios?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 ### Features
 - Adds option to enable retries on errors encountered by the Snowflake connector ([#14](https://github.com/dbt-labs/dbt-snowflake/issues/14))
+- Adds tests for incremental model unique key parameter ([core #4618](https://github.com/dbt-labs/dbt-core/pull/4618))
```
leahwicz, line 23: Can we move this to the `1.1.0` section? It looks like we haven't created that yet so could you add that and update the release date for `1.0.0` please ([it should be December 3rd](https://github.com/dbt-labs/dbt-snowflake/releases/tag/v1.0.0))?

VersusFacit: Certainly will do both here.

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
ChenyuLInx: I think this is good enough!

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
ChenyuLInx: In my mind, the count being returned is more like a side effect. The function you call in this function is actually `run_dbt` which is named based on the actual thing that happened. But you did a `len()` operation on the result returned. So I guess you can go both ways here. 
But I would avoid abstracting this and the bottom one being a function since they are not being used in multiple places, there are more things going on here than just the function name indicates, and reading the code itself is not hard. `setup_test` is a perfect name for all of the things happening

tests/integration/incremental_unique_id_test/test_incremental_unique_id.py
ChenyuLInx: Sounds good to me. Naming is hard

ChenyuLInx (approved):
Looks great to me!

PR #109, opened 2022-03-14
Add query_id to adapter payload and tests.

### LINKED ISSUE #7: Include the snowflake query id in the run results output

VersusFacit:
completes #40 
resolves #7 

### Description

See PR for more details. I've optimized the types to avoid false-positives due to truthi-strings. I pushed this over the finish line with all credit given to those behind the code since some users have been asking for this in GitHub and Slack.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 ### Fixes
 - Add unique\_id field to docs generation test catalogs; a follow-on PR to core PR ([#4168](https://github.com/dbt-labs/dbt-core/pull/4618))
 
+### Under the hood
+- Add `query_id` for a query to `run_result.json` ([#40](https://github.com/dbt-labs/dbt-snowflake/pull/40))
+
+### Contributors
+- [@joshuataylor](https://github.com/joshuataylor) ([#40](https://github.com/dbt-labs/dbt-snowflake/pull/40))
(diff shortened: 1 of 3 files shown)


jtcohen6 (approved):
LGTM! Thanks for taking this one over the finish line :)

PR #115, opened 2022-03-21
added macro for wh selection

### LINKED ISSUE #103: [CT-333] Macro for warehouse selection
### Describe the feature
It would be nice to have a macro similar to `generate_schema_name` for the selection of the snowflake warehouse to use. For example for change the warehouse depending on the environment.

I tried to do a pre hook with the `use warehouse` query but don't gives the flexibility needed, for example if you use a wildcard in the name all the queries that don't use the pre hook will fail.

### Additional context
Please include any other relevant context here.

### Who will this benefit?
Every one that wants more control and flexibility on the warehouse selected.

Yes, I have some idea about how to do it. I've performed some local tests and it's working. With some help as it's my first contribution I will be able to raise a PR.

javiCalvo:
resolves #103 Macro for warehouse selection

### Description

The change is minimal to call a macro that gets the actual / default warehouse for the model and returns the same but can be overriden for more complex logics (ie. select a different warehouse depending on the environment). Test are missing as I don't know how to start with them but the change is working.

### Checklist

- [] This PR includes tests, or tests are not required/relevant for this PR


dbt/adapters/snowflake/impl.py
```diff
@@ -82,6 +85,13 @@ def _get_warehouse(self) -> str:
...
     def _use_warehouse(self, warehouse: str):
         """Use the given warehouse. Quotes are never applied."""
+        kwargs = {
+            'warehouse': warehouse
+        }
+        warehouse = self.execute_macro(
+            SNOWFLAKE_WAREHOUSE_MACRO_NAME,
+            kwargs=kwargs
+        )
         self.execute('use warehouse {}'.format(warehouse))
```
jtcohen6, line 95: Rather than a macro that returns the name of the warehouse, and then templating + executing `use warehouse ...` right from within the Python context, I think it would be better to have the Jinja macro template out the actual SQL we want to run (`use warehouse {{ warehouse }}`) _and_ execute that SQL, via `statement` or `run_query`

That way, `self.execute_macro` accomplishes the full intended functionality, and we don't need the implicit call to `self.execute` (a.k.a. `adapter.execute`).

The macro could be named `use_warehouse` or `use_snowflake_warehouse`. To accomplish your intended functionality, you could still override that macro, adding complex conditional rules to determine the warehouse name, as long as the macro still templates out `use warehouse ...` and executes that SQL.

jtcohen6 (commented):
@javiCalvo Looping back after a long delay — let us know if this is something you'd still be interested in contributing. My interest here is around making the `_use_warehouse` behavior a bit more consistent with how other method-macro handoffs happen for dbt adapters.

dbt/adapters/snowflake/impl.py
jtcohen6: @javiCalvo Sorry for the delayed response on my end!

Fair point. It is in closer keeping with how other adapter methods/macros work in dbt today, but I also appreciate wanting a narrower macro that _just_ templates the desired SQL.

Another way we could factor this:
```sql
{% macro get_use_warehouse_sql(warehouse) -%}
    use warehouse {{ warehouse }}
{%- endmacro %}
```
```python
    def _use_warehouse(self, warehouse: str):
        """Use the given warehouse. Quotes are never applied."""
        kwargs = {
            'warehouse': warehouse
        }
        use_warehouse_sql = self.execute_macro(
            SNOWFLAKE_USE_WAREHOUSE_MACRO_NAME,
            kwargs=kwargs
        )
        self.execute(use_warehouse_sql)
```
Then your override can be just this (returning a string, no `call`, much easier to test):
```sql
{% macro get_use_warehouse_sql(warehouse) -%}
   {% set warehouse = warehouse | replace('_X_', '_' + {'stg': 'D', 'tst': 'Q', 'prd': 'P'}[target.name] + '_') %}
    use warehouse {{ warehouse }}
{%- endmacro %}
```

McKnight-42:
@javiCalvo this is looking very good, might need you to rerun pre-commit commands locally tried to fix a conflict we were having between main and guess pre-commit didn't like that (so sorry about that).

McKnight-42:
@javiCalvo going to close this in favor of #503 to get your change ready for merger thank you for all your work.

PR #132, opened 2022-04-06
add set/unset_query_tag to the dbt macro namespace

### LINKED ISSUE #133: Add `set_query_tag` to the dbt macro namespace
### Describe the feature
`set_query_tag` can only be overridden in the root project. By adding it to the dbt macro namespace, we can then leverage the dispatch functionality and override it with custom query tags in dbt packages.

The only way to achieve this is to add to the dbt macro namespace. 

### Additional context
[Please include any other relevant context here.
](https://getdbt.slack.com/archives/C37J8BQEL/p1633519489312300?thread_ts=1633377998.228400&cid=C37J8BQEL)

### Who will this benefit?
All `dbt-snowflake` users who have a central dbt package and would like to leverage custom query tags to be generated and stored at runtime. For example:

```
{
    "invocation_id": "{{ invocation_id }}",
    "run_started_at": "{{ run_started_at }}",
    "dbt_version": "{{ dbt_version }}",
    "project_name": "{{ project_name }}",
    "model":
    {
        "owners": "{{ tojson(model.meta.owners|default([])) }}",
        "database": "{{ this.database }}",
        "schema": "{{ this.schema }}",
        "name": "{{ this.name }}",
        "identifier": "{{ this.identifier }}",
        "path": "{{ model.path}}",
        "unique_id": "{{ model.unique_id }}",
        "materialized": "{{ model.config.materialized }}"
    }
}
```

Yes. PR-132 is currently in draft pending this issue creation.

robscriva:
resolves #133

### Description
There is no way of overriding the `set_query_tag` and `unset_query_tag` macros to create a custom query tag within an external package. This is because these macros aren't part of the dbt namespace and therefore cannot be overridden with dispatching precedence in the project. This change adds both macros to the dbt macro namespace so they can be overridden externally.

Discussion here: https://getdbt.slack.com/archives/C37J8BQEL/p1633519489312300?thread_ts=1633377998.228400&cid=C37J8BQEL

### Checklist


dbt/include/snowflake/macros/adapters.sql
```diff
@@ -193,6 +193,11 @@
 
 
 {% macro set_query_tag() -%}
+    {{ return(adapter.dispatch('set_query_tag', 'dbt')()) }}
+{% endmacro %}
+
+
+{% macro default__set_query_tag() -%}
```
jtcohen6, line 200: Nit: I feel like these should be `snowflake__`, rather than `default__`.

This makes no functional difference today, but in a world where another database comes along that _also_ supports "query tags," and we wanted to move the dispatched `set_query_tag` macro into `dbt-core` — the macro here would want to be `snowflake__set_query_tag()`, and the macro in `otherdb` would want to be `otherdb__set_query_tag()`.

Do you buy it?

jtcohen6 (commented):
@robscriva Looking good! One comment from me. Otherwise, this will be good to ship in v1.2.

dbt/include/snowflake/macros/adapters.sql
robscriva: SGTM. Resolved in [bcf9695](https://github.com/dbt-labs/dbt-snowflake/pull/132/commits/bcf96953cd552b74d506647154f7fa68cbd3a1bf)

jtcohen6:
Not sure what's going on with those test failures, will need to take a more detailed look

robscriva:
Just a coincidence after I swapped to use the `snowflake__` prefix right? 😂

jtcohen6:
@robscriva Definitely a coincidence. We're seeing those tests fail in the `main` branch as well, on nightly runs (e.g. [this one](https://github.com/dbt-labs/dbt-snowflake/actions/runs/2256130509)). Sorry that your PR is being held up in the meantime.

PR #158, opened 2022-05-09
Remove simple snapshot tests.

VersusFacit:
Part 1 of fixing issue CT-600. 

### Description

After not finding a quick fix and in the interest of long-term optimization, I propose following through on Jerco's call to remove this test. After we remove these tests, we can reinvite them into this repo via the adapter zone.

In a "part 2 PR," I'll make sure the following tests are either accounted for in dbt core. Alternatively, they will be readded with the new framework in a so-called "part 3 PR" back into this repo:
- tests/integration/simple_snapshot_test/check-snapshots-query-tag-expected/check_query_tag.sql
- tests/integration/simple_snapshot_test/invalidate_snowflake.sql
- tests/integration/simple_snapshot_test/test_simple_snapshot.py::TestInvalidNamespacedCustomSnapshotFiles

I traced through every file removed and looked for matches in existing dbt-core code.

### Checklist


tests/functional/adapter/test_basic.py
```diff
@@ -41,8 +41,8 @@ class TestGenericTestsSnowflake(BaseGenericTests):
     pass
 
 
-class TestSnapshotCheckColsSnowflake(BaseSnapshotCheckCols):
-    pass
+#class TestSnapshotCheckColsSnowflake(BaseSnapshotCheckCols):
+#    pass
```
jtcohen6, line 45: Ah, ok, this is concerning!

I took a deeper dive in, and I see that the snapshots are always running with `where TRUE` as their `row_changed_expr` [here](https://github.com/dbt-labs/dbt-core/blob/ab0c3510ebe70a67fb669b88e6ea36aff2f46339/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql#L149-L151). That's not right! For `check`-strategy snasphots, this variable should only return `true` if it's `check_cols = 'all'` and a new column was added—neither of which are true for the failing snapshots.

I took a quick look, and the last PR to touch this code was indeed 12 days ago: https://github.com/dbt-labs/dbt-core/pull/4893

We've got a lead, gumshoe!

jtcohen6: Updates:
- Issue: https://github.com/dbt-labs/dbt-core/issues/5222
- Proposed fix: https://github.com/dbt-labs/dbt-core/pull/5223

leahwicz:
@VersusFacit is this the PR we can close out now as not needed?

Issue #168, opened 2022-06-02
Add Grants to Materializations


nathaniel-may:
blocked by: dbt-labs/dbt-core#5263

Once the above work is merged into core, there are some tweaks that need to be made in the Snowflake adapter to fully expose the feature:
    - [ ] add a call to `apply_grants(..., revoke=True)` to all custom materializations
    - [ ] in view/table materializations, call apply_grants(..., revoke=False) instead for a performance improvement

jtcohen6:
- `get_grant_sql`: https://docs.snowflake.com/en/sql-reference/sql/grant-privilege.html (I think the default should work)
- `get_revoke_sql`: https://docs.snowflake.com/en/sql-reference/sql/revoke-privilege.html (I think the default should work)
- `get_show_grants_sql`: https://docs.snowflake.com/en/sql-reference/sql/show-grants.html (different from PostgreSQL — but we might want to make this the default?)

PR #178, opened 2022-06-27
Ct 716 add materialization

McKnight-42:
resolves: [CT-716](https://github.com/dbt-labs/dbt-snowflake/issues/168)
related to: [CT-660](https://github.com/dbt-labs/dbt-core/issues/5263)

### Before merge
- [x] Reinstate original version of `dev-requirements.txt`

### Description

implementing `apply_grants`, will add tests, for snowflake instances.

### Checklist


McKnight-42:
7/5/22 EOD REPORT: main report can be found on [SQL GRANTS](https://github.com/dbt-labs/dbt-core/pull/5369). did have a question on if the [`OWNERSHIP`](https://github.com/dbt-labs/dbt-snowflake/pull/178/files#diff-3b8dfc96ca09cc82325269902a1353fa14e4503308502ba4c64f24e037734bdbR170) check here was unneeded

dbt/adapters/snowflake/impl.py
```diff
@@ -158,5 +160,19 @@ def quote_seed_column(self, column: str, quote_config: Optional[bool]) -> str:
...
             return column
 
+    @available
+    def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
+        grants_dict = {}
+        
+        for row in grants_table:
+            grantee = row['grantee_name']
+            privilege = row['privilege']
+            if privilege != 'OWNERSHIP':
```
jtcohen6, line 170: The `OWNERSHIP` filter makes sense to me, and is one valid way of proceeding here.

My inclination would be to generalize this beyond just `OWNERSHIP`: all grants given BY the current user (=role), and NOT TO the current user (=role):
```python
for row in grants_table:

    # three relevant columns
    grantee = row['grantee_name']
    privilege = row['privilege']
    granted_by = row['granted_by']

    # super gross, but it works
    # (this is the value from profiles.yml)
    current_role = self.connections.profile.credentials.role
    
    if (
        # granted BY the current role
        granted_by.lower() == current_role.lower() and
        
        # NOT granted TO the current role
        grantee.lower() != current_role.lower()
    ):
        ...
```

jtcohen6 (commented):
Nice work!! This is running for me locally.

SHOULD fix (inefficient but correct):
-  [ ] Let's figure out the right approach to `should_revoke` on Snowflake. I left a big comment talking through this. It will have implications for the implementation we ultimately land on in `dbt-core`.

dbt/include/snowflake/macros/materializations/seed.sql
```diff
@@ -38,9 +38,11 @@
 
 {% materialization seed, adapter='snowflake' %}
     {% set original_query_tag = set_query_tag() %}
+    {% set  grant_config = config.get('grants') %}
 
     {% set relations = materialization_seed_default() %}
 
+    {% do apply_grants(relation, grant_config, should_revoke=True) %}
```
jtcohen6, line 45: This will be handled within `materialization_seed_default`, so it's not needed here

dbt/include/snowflake/macros/materializations/snapshot.sql
```diff
@@ -1,8 +1,9 @@
 {% materialization snapshot, adapter='snowflake' %}
     {% set original_query_tag = set_query_tag() %}
+    {% set  grant_config = config.get('grants') %}
 
     {% set relations = materialization_snapshot_default() %}
-
+    {% do apply_grants(target_relation, grant_config, should_revoke=True) %}
```
jtcohen6, line 6: Ditto - this will be handled within `materialization_snapshot_default`, so not needed here

jtcohen6:
@McKnight-42 @dbeatty10 Follow-ups from pairing work: https://github.com/dbt-labs/dbt-snowflake/compare/ct-716-add-materialization...jerco/grants-over-finish-line

More context in https://github.com/dbt-labs/dbt-core/pull/5369#issuecomment-1176568729. Implements most of the SHOULD + COULD comments from above.

jtcohen6:
@McKnight-42 Functional/integration tests are failing because this branch depends on the changes in https://github.com/dbt-labs/dbt-core/pull/5369. Could you update these lines:
https://github.com/dbt-labs/dbt-snowflake/blob/0203a1617585f05b17891e35e5749c736527716b/dev-requirements.txt#L3-L4
To point to your branch (`ct-660-grant-sql`):
```
git+https://github.com/dbt-labs/dbt-core.git@ct-660-grant-sql#egg=dbt-core@&subdirectory=core
git+https://github.com/dbt-labs/dbt-core.git@ct-660-grant-sql#egg=dbt-tests-adapter&subdirectory=tests/adapter
```

This is exactly how @dbeatty10 and I were doing the lift-and-shift of cross-database macros. Once the `dbt-core` PR is merged, we can repoint to the `main` branch. But it's important to know ahead of time that the functionality will work in all our adapters.

dbt/adapters/snowflake/impl.py
```diff
@@ -158,5 +159,31 @@ def quote_seed_column(self, column: str, quote_config: Optional[bool]) -> str:
...
+        grants_dict = {}
+
+        for row in grants_table:
+            grantee = row['grantee_name']
+            privilege = row['privilege']
+            granted_by = row['granted_by']
+
+            # super gross, but it works
+            # (this is the value from profiles.yml)
+            current_role = self.connections.profile.credentials.role
```
jtcohen6, line 173: TODO:
- MUST: I've remembered that `role` is technically not required in `profiles.yml`. If not specified, Snowflake will opt for the user's configured `default_role` when they connect. So, if this value is missing, we may need to run `select current_role` after all. I think our best bet would be to do that once at the start of the invocation, rather than doing it over and over.
- SHOULD: This should really be a new adapter method, rather than reaching deep in as we're doing here. That method would also allow us to abstract over whether `credentials.role` is provided, or needs to be accessed by introspective database query.

jtcohen6: Another option (!!): https://docs.snowflake.com/en/sql-reference/functions/result_scan.html

This would let us do it all in SQL — though a bit weird, to be honest, since we'd need to pass it a specific `query_id` after running the first query. We can get that `query_id`, though. Worth a shot...?

To be honest, if this proves too tricky to get this working, we can revert to the previous (simpler) logic that just filtered out any grants of the `OWNERSHIP` privilege, and treated all the rest as fair game.

jtcohen6: @dbeatty10 Interested in your gut reaction between those options. (No need for full investigation.) It does feel like the simplest thing here is reverting to your + Matt's previous logic, which just filters out the `OWNERSHIP` privilege.

McKnight-42: will revert, and push up...DONE

tests/functional/adapter/test_grants.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+from dbt.tests.adapter.grants.test_models import TestModelGrants
+
+class TestModelGrantsSnowflake(TestModelGrants):
+    pass
```
dbeatty10, line 6: A few formatting nits -- `pytest` was unused, there should be two spaces between imports and the following code, and newline at the end.

In case our automated processes didn't catch them, here they are:
```suggestion
from dbt.tests.adapter.grants.test_models import TestModelGrants


class TestModelGrantsSnowflake(TestModelGrants):
    pass

```

McKnight-42: Done

dbeatty10 (approved):
Awesome work @McKnight-42 !

PR #181, opened 2022-06-28
Ft Add materialization for Snowflake materialized view object type

leahtessem:
Resolves [#1162](https://github.com/dbt-labs/dbt-core/issues/1162)
This is an issue opened for dbt-core repository, but because the code I have written is specific to (and only tested on Snowflake), I feel that it fits in this repository. This PR (or parts of it) can be adapted and moved to live in `dbt-core` if desired.

### Description
* Adding materialization for Snowflake materialized views
  * including support for `cluster_by` config entry
  * Adding `SnowflakeRelationType` with materialized view type (including space)
  * add methods to `SnowflakeRelation` to determine if relation is MV
  * add get relation type method that includes new MV addition
* Update `snowflake__list_relations_without_caching` macro to include materialized views
  * MVs do not return from `show terse objects` so it now has to be two calls (one for tables & one for views)
* Also added macro for `create or replace`-ing materialized view similar to how the base view macro was built out in `dbt-core`

Handling switching between view types

* Add handling in view materialization to drop the object if a materialized view exists with the desired view name (similar to how tables are handled in the [create_or_replace_view](https://github.com/dbt-labs/dbt-core/blob/d5608dca322c63f1a894174b1f280ae55a217ab8/core/dbt/include/global_project/macros/materializations/models/view/create_or_replace_view.sql) macro in `dbt-core`)
* Add onto incremental materialization to drop the MV if one exists with the desired incremental table name

Added MVs in the switching relation type tests as well as their own test file

### Checklist


leahtessem:
Hi @dbt-labs/oss-maintainers, hoping to get a review here. I would like feedback on whether any of these changes should go to `dbt-core` instead and making sure that my tests are sufficient.

Getting this change merged in would allow my organization to have all of our snowflake objects managed in one place. Currently, our materialized views are managed in a way that is disjoint and prone to failure. It would also address issue https://github.com/dbt-labs/dbt-core/issues/1162 and add the feature that has been desired since 2018.

Thanks!

jtcohen6:
@leahtessem Thanks for the PR, and sorry for the delay!

I owe you a response here. To make a long story short, I think we're unlikely to merge this into `dbt-core` / adapters for native support — at least for now. For the last few years, we've kept "experimental" implementations of materialized views in this repo: https://github.com/dbt-labs/dbt-labs-experimental-features/tree/main/materialized-views

I'd like to say more about that, though. It's always worth reexamining our long-standing assumptions. I'll plan to swing back in the next few days with a fuller explanation.

Fleid:
Closing as my current thinking is that we won't support Snowflake's Materialized Views, but [Dynamic Tables](https://www.snowflake.com/blog/dynamic-tables-delivering-declarative-streaming-data-pipelines/) instead.

Thanks a lot for your contribution, it will inspire our work going forward on that topic.

PR #182, opened 2022-06-29
Feature/python model v1

ChenyuLInx:
This change currently includes table materialization and incremental materialization.

Also super happy to hear any feedback and what you think we missed.

TODOs:
- [x] functional test for python incremental model
- [x] separate out copy grants to make it work for both python and SQL model: this is not easy, created separate issue #214
- [x] add how to run python model in notebook/snowsight in comment


tests/functional/adapter/test_python_model.py
```diff
@@ -0,0 +1,69 @@
...
+    dbt.config(
+        materialized='table',
+    )
+    data = [[1,2]] * 10
+    return pandas.DataFrame(data, columns=['test', 'test2'])
+"""
+models__simple_python_model_v2 = """
+import pandas
+
+def model(dbt, session):
```
ChenyuLInx, line 49: What the model code looks like, we are enforcing session here because it is local to the function and certain operations depends on it

ChenyuLInx:
@jtcohen6  this is ready minus the part copy grants won't work for python model, it need some refactor that probably should be ticketed work. Created #214 to track it

gshank (approved):
Looks good!  Remember to remove branch from dev-requirements.txt before merging.

dev-requirements.txt
```diff
@@ -1,8 +1,7 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=tests/adapter
+git+https://github.com/dbt-labs/dbt-core.git@feature/python-model-v1#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt-core.git@feature/python-model-v1#egg=dbt-tests-adapter&subdirectory=tests/adapter
```
ChenyuLInx, line 4: remove before merging

tests/functional/adapter/test_python_model.py
```diff
@@ -0,0 +1,92 @@
...
+    dbt.config(
+        materialized='table',
+    )
+    data = [[1,2]] * 10
+    return pandas.DataFrame(data, columns=['test1', 'test3'])
```
ChenyuLInx, line 26: This failed the new return check in core, need to update that check

PR #185, opened 2022-07-01
Fixing broken 1.0.latest CI 

leahwicz:
### Description
CI is broken for `1.0.latest` because of some recent changes made in other branches but not here (i.e. change to use `python -m pip`, point to correct `dbt-core` branch, etc.). Adding these updates here to fix CI but nothing else is changing

### Checklist

- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -153,9 +153,9 @@ jobs:
 
       - name: Install python dependencies
         run: |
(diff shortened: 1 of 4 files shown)


nathaniel-may (approved):
yeah this is the right way to do this.

PR #197, opened 2022-07-13
add test for incremental model

ChenyuLInx:
Adding test for incremental python model
This test works fine locally, just didn't change the core version so everything is failing now

### DIFF
--- a/tests/functional/adapter/test_python_model.py
+++ b/tests/functional/adapter/test_python_model.py
@@ -54,6 +54,8 @@ def model(dbt, session):
     return pandas.DataFrame(data, columns=['test1', 'test3'])
 """
 
+
+
 class TestChangingSchemaSnowflake:
     @pytest.fixture(scope="class")
     def models(self):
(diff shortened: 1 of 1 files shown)


tests/functional/adapter/test_python_model.py
```diff
@@ -65,5 +67,47 @@ def test_changing_schema(self,project):
...
+select 5 as id
+"""
+
+incremental_python = """
+def model(dbt, session):
+    dbt.config(materialized="incremental", unique_key='id')
+    df = dbt.ref("m_1")
+    if dbt.is_incremental:
+        # incremental runs should only apply to 
+        df = df.filter(df.id >= session.sql(f"select max(id) from {dbt.this}").collect()[0][0])
```
ChenyuLInx, line 86: @jtcohen6  I couldn't figure out a better way to do this with a dataframe. Probably something that we can surface with the snowpark team

ChenyuLInx: I don't really like this but also couldn't figure out a way to get around it for now even with `max`. I feel okay leaving it like this for now

PR #200, opened 2022-07-18
Set query tag on the connection instead of setting it

### LINKED ISSUE #199: [CT-853] Query tag should be set on the session level and not unset for global query tags
### Describe the feature
When setting the `query_tag` this is constantly set/unset between queries, slowing down the time for runs on high latency.

This is the case when putting `+query_tag`, which is then set/unset. This is good, this is how it should be if connections aren't being remade (ie connection reuse).

However this should be set on the `session_parameters` level when it's passed in via a dbt `profiles.yml` level, where it has to execute another query. I'm 250ms-280ms away from us-east-1, so this has to do another connection per one of these, which drastically slows down the queries.
image

Living with it. :(

### Additional context
This does a lot of queries, see:
image

### Who will this benefit?
People with slow connections, or dbt with Snowflake just being faster in general.

yes

joshuataylor:
resolves #199

### Description

When setting the query_tag this is constantly set/unset between queries, slowing down the time for runs on high latency.

This is the case when putting +query_tag, which is then set/unset. This is good, this is how it should be if connections aren't being remade (ie connection reuse).

However this should be set on the session_parameters level when it's passed in via a dbt profiles.yml level, where it has to execute another query. I'm 250ms-280ms away from us-east-1, so this has to do another connection per one of these, which drastically slows down the queries.

When 
### Checklist

Unsure of this one, how to test?

Will do once approved :)


McKnight-42:
Hi @joshuataylor was taking a look at the tests that are currently failing and found that in regards to your changes we will need to add references to the `session_parameters` in the [`mock.call`](https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/unit/test_snowflake_adapter.py#L312-L319) in several spots in the [test_snowflake_adapter](https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/unit/test_snowflake_adapter.py) file.

Also had a thought that it might be worthwhile to if your up to it add a test in [test_query_tags](https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/integration/query_tag_tests/test_query_tags.py) to test setting `query_tag` in `profiles.yml` which we don't actually currently have we do test it against models, seeds and snapshots currently.

McKnight-42:
@joshuataylor looks like we have some new conflicts based on a change to our connections methods setup in part of some retry logic reconfiguration https://github.com/dbt-labs/dbt-snowflake/pull/207/files. Please let me know if you don't have time to continue on this pr.

joshuataylor:
Ah sorry, I missed your other message, i'll take a look into the comments. 👐

McKnight-42:
no worries! @joshuataylor.
I had some time today and was playing around with trying to add your changes to the new connect method style and if it helps or gives any inspiration this seems to pass current integration and functional testing for me locally. would love your thoughts? 

```@classmethod
    def open(cls, connection):
        if connection.state == "open":
            logger.debug("Connection is already open, skipping open.")
            return connection

        creds = connection.credentials
        timeout = creds.connect_timeout

        def connect():
            session_parameters = {}

            if creds.query_tag:
                session_parameters.update({"QUERY_TAG": creds.query_tag})

            handle = snowflake.connector.connect(
                account=creds.account,
                user=creds.user,
                database=creds.database,
                schema=creds.schema,
                warehouse=creds.warehouse,
                role=creds.role,
                autocommit=True,
                client_session_keep_alive=creds.client_session_keep_alive,
                application="dbt",
                session_parameters=session_parameters,
                insecure_mode=creds.insecure_mode,
                **creds.auth_args(),
            )

            return handle

 ```

**Comment by joshuataylor** [comment]
I am so sorry for the delay on this, it's been crazy lately.

I've got a super long weekend coming up now where I'm looking into this and the other PR.

The connection rework looks great 😍

**Comment by joshuataylor** [comment]
I've updated the branch to reflect the latest changes. I've also added a unit test for the query tag.

Unsure of how to do the second point around the profiles, but that is a very good point. I also couldn't get the profile test to run locally:
```bash
python -m pytest tests/functional/query_tag/test_query_tags.py
```

Just seems to sit and do nothing, but `make test` runs everything (in ~1min), so I'm guessing make test should cover that?

McKnight-42:
@joshuataylor great work love new unit tests. and no worries will do what I can to help out. I think you should be able to update the profiles.yml for testing default by using the `profiles_config_update` project method in the new functional test framework seen [here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/tests/fixtures/project.py#L134) and a example of usage in a few places in dbt-core like this [test](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/context_methods/test_env_vars.py#L68).

as for `python -m pytest tests/functional/query_tag/test_query_tags.py` command not working for you locally I might suggest rebuilding your local env and running dev-requirements again to install dependencies, remade your changes locally for myself last night and ran the query_tag functional tests and the command seemed to work for me. will find some time to try and dig deeper if this persists though.

McKnight-42:
hey @joshuataylor had some time open up so, taking this fantastic work and creating a new pr #294 to try and get that profiles test added so we can get this in. would love for you to keep informed and feel free to make suggestions on it, going to close this one out and move the conversation over there. thank you so much for all the work you've put into this

PR #202, opened 2022-07-19
[Pytest conversion] test_warehouses

McKnight-42:
resolves: continued pytest conversions.

### Description

migrating snowflake specific integration test to a new functional pytest

### Checklist


tests/functional/warehouse_test/fixtures.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+
+@pytest.fixture(scope="class")
+def models():
+    return {
+        "override_warehouse.sql": models__override_warehouse_sql,
+        "expected_warehouse.sql": models__expected_warehouse_sql,
+        "invalid_warehouse.sql": models__invalid_warehouse_sql,
+    }
```
emmyoop, line 42: This should actually be defined in your `test_warehouses.py`, not here.  That way we can easily see what `sql` files make up the models.  Same for the `project_config_models`

tests/functional/warehouse_test/fixtures.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture(scope="class")
+def project_files(
+    project_root,
+    models,
+    project_config_models,
+):
+    write_project_files(project_root, "models", models)
+    write_project_files(project_root, "project-config-models", project_config_models)
```
emmyoop, line 60: This is also unnecessary when you define the models in your tests.  This is done when we're setting up the project(https://github.com/dbt-labs/dbt-core/blob/1071a4681df91633301fdf23e34de819b66fbeb7/core/dbt/tests/fixtures/project.py#L331-L338).

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+
+
+class TestModelWarehouse():
+    def dir(value):
+        return os.path.normpath(value)
+
+    @pytest.fixture(scope="class")
+    def model_path(self):
+        return self.dir('models')
```
emmyoop, line 20: Is this pattern something that needs to be done for snowflake?  It's just not a pattern I've seen in our tests yet but I haven't worked with the snowflake tests either.

emmyoop (commented):
I would remove the fixture file completely and just put it all into your test file.  It's the pattern most of the tests follow for fixtures that are not shared across files.

ChenyuLInx:
The fixture thing is probably the fault of my conversion script, since I just did it as a standard thing to do. Good call out @emmyoop

dbeatty10 (commented):
I'm probably overlooking this somehow -- how is this test only running against the Snowflake test profile and not running against other database adapters?

McKnight-42:
> 
@dbeatty10 there are a few integration/functional tests that fall under very niche use cases so we have some tests that only appear in the adapters that need them. this being one of them that is literally only used for snowflake so there is no dbt-core version of it to be inherited (is this something we should move away from, who knows).

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+
+
+@pytest.fixture(scope="class")
+def models():
+    return {
+        "override_warehouse.sql": models__override_warehouse_sql,
+        "expected_warehouse.sql": models__expected_warehouse_sql,
+        "invalid_warehouse.sql": models__invalid_warehouse_sql,
+    }
```
emmyoop, line 47: Move this inside `TestModelWarehouse`

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@pytest.fixture(scope="class")
+def project_config_models():
+    return {
+        "override_warehouse.sql": project_config_models__override_warehouse_sql,
+        "expected_warehouse.sql": project_config_models__expected_warehouse_sql,
+    }
```
emmyoop, line 55: Move this inside `TestConfigWarehouse`.  It makes the discoverability of what models are used in what tests much higher.

emmyoop (changes requested):

tests/functional/warehouse_test/test_warehouses.py
McKnight-42: done, had to rename `project_configt_models` to `models` once I moved it into the class but still passes locally

tests/functional/warehouse_test/test_warehouses.py
McKnight-42: done!

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -0,0 +1,84 @@
...
+
+class TestModelWarehouse:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "override_warehouse.sql": models__override_warehouse_sql,
+            "expected_warehouse.sql": models__expected_warehouse_sql,
+            "invalid_warehouse.sql": models__invalid_warehouse_sql,
+        }
+    def test_snowflake_override_ok(self, project):
```
emmyoop, line 49:
```suggestion
        }

    def test_snowflake_override_ok(self, project):
```

emmyoop (approved):
Just a tiny formatting suggestions then this looks good to merge in!

PR #203, opened 2022-07-20
Add release_connection setting to not release the connection

### LINKED ISSUE #201: [CT-854] Cache authentication (token) for subsequent connections
### Describe the bug
With Snowflake, logging in is slow (as it should be, given password hashing etc), so it should be avoided.

Every query that is executed causes a new login, as the connection is marked as closed. This should not happen, the connection should be marked as open so you're not forced to login again.

### Steps To Reproduce
1. Have multiple queries, or set the threads to a low limit
2. Issue the queries
3. See multiple logins, you can verify by:
```sql
select *
from table(information_schema.login_history_by_user())
order by event_timestamp desc;
```

### Expected behavior
A single login happens, maybe by thread?

### System information
Happens on MacOS/Linux, pretty sure it's not a OS issue..
Python 3.10.4, dbt 1.1.1

joshuataylor:
resolves #201

### Description
When a connection is released, it is marked as closed. This is fine for other connectors, but Snowflake is a Cloud Data Warehouse where you could have a high latency connection to the server. You are also logged in for every query you issue with dbt, which can cause query times to be slow, as you have to go through the entire login process again.

Instead, this will allow you to reuse the connection.

Now this makes a MASSIVE difference for me with views, not so much with tables, but will still take a second or so off table creation as you don't have to reauthenticate. Views that were taking 2-3 seconds now take <0.5s! I'm from Perth, Australia, so your numbers will be a lot better than mine, as I'm ~250ms away from us-east-1.

Using 16 threads:

Before:  Finished running 101 view models, 49 table models in 0 hours 1 minutes and 34.24 seconds (94.24s).
Now: Finished running 101 view models, 49 table models in 0 hours 1 minutes and 6.09 seconds (66.09s).

Just views:
Before: Finished running 148 view models in 0 hours 0 minutes and 48.72 seconds (48.72s).
Now: Finished running 148 view models in 0 hours 0 minutes and 12.05 seconds (12.05s).

The biggest win for this comes from `dbt test`:

dbt test (8 threads):
Before: Finished running 798 tests in 0 hours 5 minutes and 58.47 seconds (358.47s).
Now: Finished running 798 tests in 0 hours 0 minutes and 56.99 seconds (56.99s).

Before adding tests around this, I would like to know that everyone is happy with this change. I've put this behind a `release_connection` setting, which defaults to `True`. This will allow users to test this and ensure that it works for them, then we could make this default at some point.

### Checklist


jtcohen6:
Super cool findings @joshuataylor! I'm going to mark this ready for technical review.

I do think we'll need/want extensive testing for this. A few years ago, we ran into issues around leaving connections open + `client_session_keepalive` (https://github.com/dbt-labs/dbt-core/issues/2645, https://github.com/dbt-labs/dbt-core/pull/2650 — I linked these in the `dbt-core` issue as well):
> The problem here is that the background thread that the snowflake connector starts is hanging, because dbt never calls `close()` on the handle.

So - worth thoroughly testing, including that specific reproduction case in the RPC server / IDE!

joshuataylor:
Yeah that might leave open connections for RPC. Could we maybe check that if the `client_session_keepalive` is on, we won't do this?

It could cause an impact if you have say 8 threads, then 1 long running model as the final model, or if you have a model in the middle which is long.

For example with this change if you had the setting enabled:

1. model1, model2, model3 are quick to execute and can run in parallel, great
2. model4 which relies on the above models takes a while to execute, oh no now we have 2 open connections sitting with the heartbeat, which `snowflake-connector-python` will keep alive for us.
3. model5-8 will reuse connections as per normal.

Few short term options:

1. We don't allow `client_session_keepalive`  and connection reuse to be enabled at the same time, or display a warning.
2. Actually close any extra threads after a time period. This change which will probably require dbt-core changes (but honestly we should just do it on the `dbt-snowflake` connector)

Spince:
I want this!

McKnight-42:
@joshuataylor seems this pr will also need the `changie` process ran, and might have some minor conflicts with more recent work in same space. based around some new connection logic changes similar to your `query_tag` pr

VersusFacit:
Josh! Hello! 👋  This is a huge win when we get it merged in. 💥 

 I've had my eyes on this for a bit, and have had some time to recreate the issue and experiment with your solution! I'm wrapping up an evaluation of your solution. My main concerns at the moment are:
* knowing precisely how these booleans affect the code
* asynchronously concerns -- I get scared when I see lock invocations

I'm doing a bit of testing and can circle back soon. If you'd like to keep working on this at that time, I'm happy to guide you through, or I could cherry pick your work so far onto a dev branch. Up to your appetite for pushing this over the finish line.

I'm hoping I can shut up my anxieties though and we can basically merge your code as is.

Also, fyi, As the synk error is not apparently your fault, and something we'll request an override for at an appropriate time!

joshuataylor:
All good! Sorry, been super busy here with a few things but can make some time to work through this here if anyone has any questions.

I've been running this locally and it's been working fine, but I also am unsure about the long running queries as mentioned. IMHO, the whole thing should be overhauled (connections are incredibly tricky, and when you have oauth and other things you might want to oauth first, then use that token elsewhere, I believe this is currently an issue with dbt-snowflake) but I'll make another discussion for that.

VersusFacit:
First of all no apologies needed! It's a busy time of year and there's certainly no immediate expectations you'd be down to deal with all those intricacies. I'm just gauging your continued interest. We ❤️  your commits  

We'd restart by getting this branch passing, since right now it's skipping some adapter tests (we'll handle the overrides for other failing checks like synk which aren't on you)

PR #207, opened 2022-07-20
connection retry refactor, add defaults with exponential backoff

### LINKED ISSUE #205: [CT-874] snowflake: standardize connection retries.
## Criteria
- default to one retry after 1 second if `connection_retries` and `connect_timeout` are not specified.
- if users specify more than one retry, use x^2 exponential backoff
- preserve existing settings. i.e. 
    + `connect_retries`
    + `connect_timeout`
    + `retry_all`
    + `retry_on_database_errors`
    + `query_tag`

nathaniel-may:
resolves #205

- Refactors retry logic to use `retry_connection` from core. 
- Preserves existing behavior, 
- Adds consistent defaults with other adpaters:
    + if `connect_retries` and `connect_timeout` are not specified: 1 retry after 1s
    + if `connect_retries` is specified but `connect_timeout` is not, it will use exponential backoff

This is a non-trivial change so I'll wait for sign off from @jtcohen6 before bothering to add the tests.

## TODO
- ~add similar tests from [postgres PR](https://github.com/dbt-labs/dbt-core/pull/5432)~ <- those are testing the library function in core so we're covered here too.


dbt/adapters/snowflake/connections.py
```diff
@@ -59,8 +59,8 @@ class SnowflakeCredentials(Credentials):
     proxy_host: Optional[str] = None
     proxy_port: Optional[int] = None
     protocol: Optional[str] = None
-    connect_retries: int = 0
-    connect_timeout: int = 10
+    connect_retries: int = 1
```
jtcohen6, line 62: 👍 

Also discussed: using `connect_retries` as the standard name everywhere (including Postgres/Redshift)

dbt/adapters/snowflake/connections.py
```diff
@@ -260,103 +260,47 @@ def open(cls, connection):
...
-        else:
-            logger.debug(
-                "Got an error when attempting to open a snowflake "
-                "connection: '{}'".format(error)
-            )
+        retryable_exceptions = []
+        if creds.retry_all:
+            retryable_exceptions = [snowflake.connector.errors.Error]
+        elif creds.retry_on_database_errors:
+            retryable_exceptions = [snowflake.connector.errors.DatabaseError]
```
jtcohen6, line 294: IMO let's use `retry_on_database_errors: True` as our default. Users can optionally turn if off, or they can opt into all errors via `retry_all`.

(@nathaniel-may may also be inspired to dig further into Snowflake docs)

dbt/adapters/snowflake/connections.py
```diff
@@ -59,8 +59,8 @@ class SnowflakeCredentials(Credentials):
     proxy_host: Optional[str] = None
     proxy_port: Optional[int] = None
     protocol: Optional[str] = None
-    connect_retries: int = 0
-    connect_timeout: int = 10
+    connect_retries: int = 1
+    connect_timeout: Optional[int] = 10
```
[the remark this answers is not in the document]
nathaniel-may: defaulted to `None` be cause once you start having defaults, you can't stop 🤪

dbt/adapters/snowflake/connections.py
nathaniel-may: I found more sensible defaults by looking at the [connector source](https://github.com/snowflakedb/snowflake-connector-python/blob/2ccd3c811958170854ff4cac159e7c125bf8b4f3/src/snowflake/connector/errors.py).

jeremyyeo:
Hey @nathaniel-may whilst you folks are working on this one... does this sync up the `connect_timeout` config of dbt-snowflake with all the likes of redshift/postgress adapter `connect_timeout`?

We recently put in a default of `connect_timeout: 30` in dbt Cloud's snowflake profile (https://dbtlabs.atlassian.net/browse/SHP-588) since this config is unavailable to end users. When looking at some of the logs once this was deployed, we realized that this doesn't behave the same way as the other adapter's config (and then I double checked the docs[^1] which does clearly state the behaviour here). As someone familiar with other adapters `connect_timeout` config, we automatically thought this config would behave the exact samesies.

This config for Snowflake's behaviour is:

> If a connection terminations, wait `connect_timeout` seconds before retrying the connection.

As opposed to Redshift/PG's:

> While connecting, wait `connect_timeout` seconds before dropping the connection.

[^1]: https://docs.getdbt.com/reference/warehouse-profiles/snowflake-profile#retry_on_database_errors

nathaniel-may:
Hey @jeremyyeo! This is part of a larger effort to keep timeout config behavior to roughly the same standard:

The goal:
- default to one retry after 1s for retryable errors
- if the user configures a number of retries, use exponential backoff for retryable errors

Exceptions:
- Snowflake already had additional retry logic so this implementation is backwards compatible for users who have that already configured. To be specific- if the user has `connect_timeout` configured, retries will use that instead of exponential backoff.
- Postgres and Redshift will attempt to retry some non-retryable errors as well due to a psycopg2 limitation.
- Spark will attempt to retry ALL errors due to a driver limitation.
- GCP will use its own exponential backoff strategy for only retryable errors as implemented in their `@Retry` decorator. The behavior is super similar but under Google's control not ours.

dbt/adapters/snowflake/connections.py
```diff
@@ -260,103 +272,57 @@ def open(cls, connection):
...
+            RequestTimeoutError,
+            BadGatewayError,
+            OtherHTTPRetryableError,
+            BindUploadError,
+        ]
+        # these two options are for backwards compatibility
+        if creds.retry_all:
+            retryable_exceptions = [Error]
+        elif creds.retry_on_database_errors:
+            retryable_exceptions.insert(0, DatabaseError)
```
jtcohen6, line 316: 👍

jtcohen6 (approved):
LGTM! Excited to get this merged. Sorry for the delay with final review

leahwicz:
@nathaniel-may can we merge this in for the end of the sprint and call it done?

PR #229, opened 2022-08-09
add pandas as a defualt package

### LINKED ISSUE #228: [CT-1017] No module named 'pandas' for python models
### Describe the bug
when running dbt python models on snowflake without using pandas, dbt will throw an error
```
13:09:39    ModuleNotFoundError: No module named 'pandas'
13:09:39     in function PYTHON_MODEL__DBT_SP with handler main
```
This is because in materialization code we use pandas to determine whether the returned dataframe is a snowpark dataframe or a pandas dataframe, and do slightly different things based on that

### Steps To Reproduce
Run a python model without specify pandas as part of the package

### Expected behavior
Everything works just fine.

### Solution
Always add pandas as the required dependencies.
This means each user will have to agree the anaconda policy on snowflake.

### Other comment
one really odd thing is that we have a [basic python model test](https://github.com/dbt-labs/dbt-core/blob/main/tests/adapter/dbt/tests/adapter/python_model/test_python_model.py#L13)(written in core, inhereted in dbt-snowflake) that run things in this way on our testing snowflake account, and doesn't run into the same issue. maybe we added `pandas` as a global library there?

ChenyuLInx:
resolves #228

### Description
Add pandas as the default packages for dbt-snowflake

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -180,7 +180,13 @@ class SnowflakeAdapter(SQLAdapter):
         database = getattr(parsed_model, "database", self.config.credentials.database)
         identifier = parsed_model["alias"]
         proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
-        packages = ["snowflake-snowpark-python"] + parsed_model["config"].get("packages", [])
+        packages = parsed_model["config"].get("packages", [])
+        # adding default packages we need to make python model work
+        default_packages = ["pandas", "snowflake-snowpark-python"]
+        package_names = [package.split("==")[0] for package in packages]
+        for defualt_package in default_packages:
+            if defualt_package not in package_names:
+                packages.append(defualt_package)
         packages = "', '".join(packages)
         python_stored_procedure = f"""
 CREATE OR REPLACE PROCEDURE {proc_name} ()


dbt/adapters/snowflake/impl.py
```diff
@@ -180,7 +180,13 @@ def submit_python_job(self, parsed_model: dict, compiled_code: str):
...
         identifier = parsed_model["alias"]
         proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
-        packages = ["snowflake-snowpark-python"] + parsed_model["config"].get("packages", [])
+        packages = parsed_model["config"].get("packages", [])
+        # adding default packages we need to make python model work
+        default_packages = ["pandas", "snowflake-snowpark-python"]
+        package_names = [package.split("==")[0] for package in packages]
+        for defualt_package in default_packages:
+            if defualt_package not in package_names:
+                packages.append(defualt_package)
```
jtcohen6, line 189:
```suggestion
        for default_package in default_packages:
            if default_package not in package_names:
                packages.append(default_package)
```

jtcohen6 (commented):
👍 code looks cleaner

Another approach we could take: Adding some `if pandas` logic within our "materialization" code, to detect if it's installed before trying to call it. Two reasons:
- Requiring `pandas` for all Python models means that enabling Anaconda packages / accepting third-party terms is a prerequisite for using Snowpark Python at all (https://github.com/dbt-labs/docs.getdbt.com/pull/1848)
- Could loading `pandas` in every stored proc risk slowing down its runtime?

Basic idea: this code

https://github.com/dbt-labs/dbt-snowflake/blob/340c3faaa43cff425c92161e4918beb171f1196f/dbt/include/snowflake/macros/materializations/table.sql#L44-L48

would become something like:

```python
try:
    import pandas
else:
    pandas = None

if pandas and isinstance(df, pandas.core.frame.DataFrame): 
     # session.write_pandas does not have overwrite function 
     df = session.createDataFrame(df)
```

ChenyuLInx:
@jtcohen6 As I looked into this more, I realized that the reason we don't run into this issue without specify `pandas` was because pandas is required(maybe optional) for snowflake-snowpark package. I can def do the checking pandas way, but also think it is okay to require it. 

Product call and we will make it happen.

Screen Shot 2022-08-15 at 5 14 01 PM
-

jtcohen6:
@ChenyuLInx Sounds good! Let's do the other approach then: check to see whether `pandas` is installed, and then proceed accordingly.

Only question: It's not totally clear to me if a user could still call `.to_pandas()` on a Snowpark data frame, even if `pandas` isn't loaded into the environment. I think this would fail:
```python
def model(dbt, session):
    df = dbt.ref('some_model')
    return df.to_pandas()
```
And so the user would need to configure it as:
```python
def model(dbt, session):
    dbt.config(packages = ['pandas'])
    df = dbt.ref('some_model')
    return df.to_pandas()
```

ChenyuLInx:
@jtcohen6 Looking though the Slack chat again of how this issue is created, here's my current theory.

1. For snowpark python run, you cannot use any third party library before accept Anaconda Terms, but you can still use `snowflake-snowpark` package, and when you specify it as a package, snowflake will install it without including the [optional](https://docs.snowflake.com/en/developer-guide/snowpark/python/setup.html#installation-instructions) pandas dependency. Thus user would run into the pandas not available error
2. Once you accept the Anaconda term, when you specify `snowflake-snowpark` as a package, snowflake will automatically install it with the optional pandas dependency. And that's why we never run into an issue with it.
3. `to_pandas` probably will not work if you don't accept the Anaconda term

So the current solution of checking whether pandas is available should work just fine. 
- before accepting the term, your pure snowpark python code works fine
- after accepting the term, you still don't need to add pandas to every environment, it will be automatically added.

Might worth confirm with the snowpark team about it since both our testing and production account have accepted the term and there's no way I can test it.

dbt/adapters/snowflake/impl.py
```diff
@@ -180,7 +180,13 @@ def submit_python_job(self, parsed_model: dict, compiled_code: str):
         database = getattr(parsed_model, "database", self.config.credentials.database)
         identifier = parsed_model["alias"]
         proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
-        packages = ["snowflake-snowpark-python"] + parsed_model["config"].get("packages", [])
+        packages = parsed_model["config"].get("packages", [])
```
ChenyuLInx, line 183: This make sure that if user also specify `snowflake-snowpark-python` in the config, we will still run everything well

gshank (approved):

PR #236, opened 2022-08-16
change on catalog.sql in case big information_schema

mauricioobgo:
resolves #

### Description

This optimize catalog.sql when query larger information_schema in order to create faster documentation in dbt

### Checklist


Fleid:
I feel like this is useful but without context it's hard to gauge.
Do we have an issue for it? Have you benchmarked the gain in performance? Do we know of any side effects?

mauricioobgo:
> I feel like this is useful but without context it's hard to gauge. Do we have an issue for it? Have you benchmarked the gain in performance? Do we know of any side effects?

- No, there is no currently an issue.
- The first cte was performing just a select on all the information schema as it is the columns cte and you only filtering at the end, hence your doing a select all of twice in the same information schema before filtering with the join but if you filter before the two cte at the end you are joining less amount of data therefore an increase in performance. This reduce or catalog to run in 30 min from 3 hrs, since our schemas have in avg 10Th tables.
- Side effects, would be not looping the entire information schema within the database just the ones you have.

.changes/unreleased/Features-20220815-203822.yaml
mikealfare, line 1: Please delete this changelog as it's a duplicate and it points to a non-existent issue.

mauricioobgo: Deleted @mikealfare
Thank you

mikealfare (approved):

PR #244, opened 2022-08-30
Conversion of snowflake ephemeral test

McKnight-42:
adapter specific conversion of [CT-1050](https://github.com/dbt-labs/dbt-core/pull/5699)

### Description

converting test_ephemeral to pytest functional test.

### Checklist


tests/functional/adapter/test_ephemeral.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+from dbt.tests.adapter.ephemeral.test_ephemeral import BaseEphemeralMulti
```
McKnight-42, line 2: Why does this not count as a proper import? at same level as all other files in the adapter section of `core`, and i'm importing this current branch i'm testing against in `dev_requirements`

tests/functional/adapter/test_ephemeral.py
```diff
@@ -0,0 +1,10 @@
...
+from dbt.tests.util import run_dbt, check_relations_equal
+
+class TestEphemeralMultiSnowflake(BaseEphemeralMulti):
+
+    def test_ephemeral_multi_snowflake(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 3
+        check_relations_equal(project.adapter, ["SEED", "DEPENDENT", "DOUBLE_DEPENDENT", "SUPER_DEPENDENT"])
```
McKnight-42, line 10: Biggest difference between the `Multi` test in` core `and here is the way the `relation` name gets uppercased

tests/functional/adapter/test_ephemeral.py
emmyoop: It looks like the new test in `dbt-core` is not nested under adapters so that path is not valid.

tests/functional/adapter/test_ephemeral.py
emmyoop: I'm assuming snowflake needs these uppercased?  Are there other tests that do something similar?  How do they handle it?

emmyoop (commented):
Just a couple questions.

tests/functional/adapter/test_ephemeral.py
McKnight-42: oh so this one is incorrect? https://github.com/dbt-labs/dbt-core/pull/5735/files created a second pr to test this attempt against

tests/functional/adapter/test_ephemeral.py
McKnight-42: `test_warehouses` has a similar part of the test here https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/functional/warehouse_test/test_warehouses.py#L80-L85 looks like the models being lowercase initially shouldn't be a issue because it compares after they have been run which would return them upper

ChenyuLInx: Or you can make the things to check(`["SEED", "DEPENDENT", "DOUBLE_DEPENDENT", "SUPER_DEPENDENT"]`) a fixture, then overwrite it in dbt-snowflake

ChenyuLInx (approved):

PR #260, opened 2022-09-14
add a way to use anonymous sproc

### LINKED ISSUE #256: [CT-1164] Snowpark: submit anon sproc instead of named sproc
The Snowpark team made an even better way to submit (anonymous) stored procedures. Hooray! I think we should use it. Only tricky thing is timing.
Copied Slack conv over
- Add another model config, use_anonymous_sproc, that will be False by default in v1.3
- Users can set the config to True for all their models, if they opt into Private Preview now, or after this feature is Open
- Preview + enabled by default for all accounts in October
- We turn it on by default in v1.4
(placeholder ticket, fill in more details later)

ChenyuLInx:
resolves #256 

### Description
support submit python model use anonymous sproc

### Checklist


gshank (approved):

.changes/unreleased/Breaking Changes-20220913-171057.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Breaking Changes
```
jtcohen6, line 1: Is this actually a breaking change?

dbt/adapters/snowflake/impl.py
```diff
@@ -179,16 +179,36 @@ def submit_python_job(self, parsed_model: dict, compiled_code: str):
         schema = parsed_model["schema"]
         database = parsed_model["database"]
         identifier = parsed_model["alias"]
-        proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
+
         packages = parsed_model["config"].get("packages", [])
+        python_version = parsed_model["config"].get("python_version", "3.8")
+        use_anonymous_sproc = parsed_model["config"].get("use_anonymous_sproc", False)
```
jtcohen6, line 185: FYI @lostmygithubaccount - this needs to be documented

dbt/adapters/snowflake/impl.py
```diff
@@ -199,16 +219,17 @@ def submit_python_job(self, parsed_model: dict, compiled_code: str):
...
-            auto_begin=False,
-            fetch=False,
-        )
+CALL {proc_name}();
+
+            """
+        response, _ = self.execute(python_stored_procedure, auto_begin=False, fetch=False)
+        if not use_anonymous_sproc:
+            self.execute(
+                f"drop procedure if exists {proc_name}(string)",
```
jtcohen6, line 229:
```suggestion
                f"drop procedure if exists {proc_name}()",
```
Jeremiah H (from Snowflake) noticed that this fails to drop the sproc, because it's created without any arguments `()` and then `drop if exists` with one argument `(string)`

jtcohen6 (commented):
@ChenyuLInx A couple comments — sorry I was late in leaving these!

ChenyuLInx:
Thanks for catching it! @jtcohen6 Opened new PR for the feedback!

PR #294, opened 2022-10-11
[CT-853] Query_Tag connection set change

### LINKED ISSUE #199: [CT-853] Query tag should be set on the session level and not unset for global query tags

McKnight-42:
resolves #199 #200

### Description

When setting the query_tag this is constantly set/unset between queries, slowing down the time for runs on high latency.

This is the case when putting +query_tag, which is then set/unset. This is good, this is how it should be if connections aren't being remade (ie connection reuse).

However this should be set on the session_parameters level when it's passed in via a dbt profiles.yml level, where it has to execute another query. I'm 250ms-280ms away from us-east-1, so this has to do another connection per one of these, which drastically slows down the queries.

When
### Checklist


dbt/adapters/snowflake/connections.py
```diff
@@ -286,6 +291,7 @@ def connect():
                 client_session_keep_alive=creds.client_session_keep_alive,
                 application="dbt",
                 insecure_mode=creds.insecure_mode,
+                session_parameters=session_parameters,
                 **creds.auth_args(),
             )
```
colin-rogers-dbt, line 297: Shouldn't we also remove the below code? 
```
if creds.query_tag:
                handle.cursor().execute(
                    ("alter session set query_tag = '{}'").format(creds.query_tag)
                )
```

McKnight-42: removed this seemingly duplicated logic and tested locally, didn't seem to cause any issues or leave leftover tests in snowflake by removing

colin-rogers-dbt (approved):
LGTM

PR #379, opened 2023-01-08
Have snowflake build incremental model optimized _tmp tables with temporary views.

### LINKED ISSUE #306: [CT-1467] [CT-1466] [Regression] v1.3: views instead of temp tables cause parallel sessions interfere: possible data leak and failure
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality

### Current Behavior

We are using Prefect to run several instances of the same dbt-snowflake project with different variables in parallel. Almost randomly, the dbt task fails with this error: 
`SQL compilation error: Object 'XYZ__DBT_TMP' does not exist or not authorized.`
Why?
The way dbt v <= 1.2 works is that it first creates a temporary table based on the dbt model like:
`Create or replace temp table XYZ__dbt_tmp`
And then merges this temporary table into the target table.
Temporary tables in snowflake are confined to the specific session in which they are created. If session 1 creates a temp table XYZ, session 2 can’t see temp table XYZ, so theoretically, infinite temp tables with the same name can coexist in parallel sessions without interfering with each other. When the session ends, the temporary table gets destroyed.
However, dbt v1.3 has replaced temporary tables with views. So it first creates a view like:
`Create or replace view XYZ__dbt_tmp`
And then merges the view into the target table, and finally drops the view like:
`Drop view if exists XYZ__dbt_tmp`
Views ARE NOT confined to sessions, so several parallel sessions trying to create views with the same name interfere with each other. This means that session 1 does `create or replace view XYZ__dbt_tmp`, session 2 does `create or replace view XYZ__dbt_tmp`, session 1 finishes the merge using the view created by session 2 and does a `drop view if exists XYZ__dbt_tmp`, and when session 1 wants to merge the XYZ__dbt_tmp view, it doesn’t exist anymore, hence we get the error 
`SQL compilation error: Object 'XYZ__DBT_TMP' does not exist or not authorized.`
We use session variables to limit sessions to specific datasets. This behavior also causes data leak.

Expected behavior is for the views to show the same behavior as temp tables (being confined to a session). For example, suffixing the view name with the session id to differentiate between the view in different sessions.

### Steps To Reproduce

This is not easy to reproduce, but if you trigger several runs of the same dbt project, and the timing is right (drop view in session 1 happens before merge in session 2), you can recreate it.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Irrelevant
- Python: 3.9
- dbt (working version): <=1.2
- dbt (regression version): 1.3
```

snowflake

### Additional Context

_No response_

VersusFacit:
resolves #306 

cc: florian since you might know more than me on our ideal use of the `create_view_as` function across adapters. It's a bit fuzzy to me, and I want to make sure my fix is suitable to long term design considerations.

### Description

> Why this? Why not build the `temporary` flag directly into the `create_view_as` function?

I did this at first, actually! It worked great....once I spent some time learning/editing the undergirding logic in dbt core.

However, this raises the questions/concerns:
* we'd alter the logic in any other adapter that implements a `create_view_as` function. This is a headache. 
* Redshift and Bigquery -- in fact, a majority of our chief cloud warehouses -- lack a `temporary` predicate for `view`s
    * from a design perspective, it strikes me that having all adapter specs to support a `temporary` predicate on `create_view_as` may not make sense to users of those platforms lacking that construct!
* the only time we employ these is on `incremental` materialized models, it seems

### Solution

Treat the generic method override as a wrapper for a flexible, `temporary`-friendly Snowflake specific function that does the heavy lifting.

#### Testing routine

This is a hard thing to test. I came up with the following.
1. run the script below and see the error in question arise (this happened against a snowflake warehouse on several of the 20 iterations)
2. run the script below to completion without error
3. check this is the correct order of predicates with both `temporary` and `secure` -- it is ✅ 

```sh
#!/bin/bash -xe

for i in {1..20}
do
  # orders is part of jaffle shop; I just made it an incremental model for this
  dbt run -s orders &
  job_one=$!
  sleep 8 # for my system, 8 was more reliable than 7 or 9
  kill -STOP $job_one

  sleep 2

  dbt run -s orders
  kill -CONT $job_one
done
```

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -72,7 +72,7 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as_with_temp_flag(temporary, relation, sql) -%}
+{% macro snowflake__create_view_as(relation, sql, temporary=False) -%}
   {%- set secure = config.get('secure', default=false) -%}
   {%- set copy_grants = config.get('copy_grants', default=false) -%}
   {%- set sql_header = config.get('sql_header', none) -%}
@@ -92,10 +92,6 @@
   {% if copy_grants -%} copy grants {%- endif %} as (
     {{ sql }}
   );
-{% endmacro %}
-
-{% macro snowflake__create_view_as(relation, sql) -%}
-  {{ snowflake__create_view_as_with_temp_flag(False, relation, sql) }}
 {% endmacro %}
 
 {% macro snowflake__get_columns_in_relation(relation) -%}
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -72,7 +72,7 @@
 )
 {% endmacro %}
 
-{% macro create_view_as_with_temp_flag(temporary, relation, sql) -%}
+{% macro snowflake__create_view_as_with_temp_flag(temporary, relation, sql) -%}
   {%- set secure = config.get('secure', default=false) -%}
   {%- set copy_grants = config.get('copy_grants', default=false) -%}
   {%- set sql_header = config.get('sql_header', none) -%}
@@ -95,7 +95,7 @@
 {% endmacro %}
 
 {% macro snowflake__create_view_as(relation, sql) -%}
-  {{ create_view_as_with_temp_flag(False, relation, sql) }}
+  {{ snowflake__create_view_as_with_temp_flag(False, relation, sql) }}
 {% endmacro %}
 
 {% macro snowflake__get_columns_in_relation(relation) -%}
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -78,9 +78,10 @@
   {%- set sql_header = config.get('sql_header', none) -%}
 
   {{ sql_header if sql_header is not none }}
+  -- CHECK ORDER
   create or replace {% if secure -%}
     secure
-  {%- endif %} view {{ relation }}
+  {%- endif %} temporary view {{ relation }}
   {% if config.persist_column_docs() -%}
     {% set model_columns = model.columns %}
     {% set query_columns = get_columns_in_query(sql) %}
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -72,7 +72,7 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as(relation, sql, temporary=False) -%}
+{% macro snowflake__create_view_as_with_temp_flag(relation, sql, temporary=False) -%}
   {%- set secure = config.get('secure', default=false) -%}
   {%- set copy_grants = config.get('copy_grants', default=false) -%}
   {%- set sql_header = config.get('sql_header', none) -%}
@@ -92,6 +92,10 @@
   {% if copy_grants -%} copy grants {%- endif %} as (
     {{ sql }}
   );
+{% endmacro %}
+
+{% macro snowflake__create_view_as(relation, sql) -%}
+  {{ snowflake__create_view_as_with_temp_flag(relation, sql) }}
 {% endmacro %}
 
 {% macro snowflake__get_columns_in_relation(relation) -%}
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -72,7 +72,7 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as_with_temp_flag(relation, sql, temporary=False) -%}
+{% macro snowflake__create_view_as_with_temp_flag(relation, sql, is_temporary=False) -%}
   {%- set secure = config.get('secure', default=false) -%}
   {%- set copy_grants = config.get('copy_grants', default=false) -%}
   {%- set sql_header = config.get('sql_header', none) -%}
@@ -80,7 +80,7 @@
   {{ sql_header if sql_header is not none }}
   create or replace {% if secure -%}
     secure
-  {%- endif %} {% if temporary -%}
+  {%- endif %} {% if is_temporary -%}
     temporary
   {%- endif %} view {{ relation }}
   {% if config.persist_column_docs() -%}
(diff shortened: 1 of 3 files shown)


dbt/include/snowflake/macros/adapters.sql
```diff
@@ -78,9 +78,10 @@
   {%- set sql_header = config.get('sql_header', none) -%}
 
   {{ sql_header if sql_header is not none }}
+  -- CHECK ORDER
   create or replace {% if secure -%}
     secure
-  {%- endif %} view {{ relation }}
+  {%- endif %} temporary view {{ relation }}
```
VersusFacit, line 84: Alternatives considered include:
1. forcing all `create_view_as` functions to take a `temporary` boolean and guard against that here.
2. duplicate this method with temporary active and call that in the incremental materialization
    * really only helps if the goal is to keep `create_view_as` unadulterated, but I saw no real need for that

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -72,14 +72,16 @@
...
+{% macro create_view_as_with_temp_flag(temporary, relation, sql) -%}
   {%- set secure = config.get('secure', default=false) -%}
   {%- set copy_grants = config.get('copy_grants', default=false) -%}
   {%- set sql_header = config.get('sql_header', none) -%}
 
   {{ sql_header if sql_header is not none }}
   create or replace {% if secure -%}
     secure
+  {%- endif %} {% if temporary -%}
+    temporary
```
VersusFacit, line 84: I _very_ specifically check that this is the correct relative order of `secure` and `temporary`. Snowflake after all has no mention of this in their main docs.

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -72,14 +72,16 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as(relation, sql) -%}
+{% macro create_view_as_with_temp_flag(temporary, relation, sql) -%}
```
nathaniel-may, line 75: Are there any backwards compatibility issues with renaming this macro? i.e. - does anyone override this in their projects?

VersusFacit: I specifically picked a name that I didn't think anyone was using anywhere. Is that too naive?

colin-rogers-dbt: I think we want a `snowflake__` prefix here to serve as a namespacing and make sure we don't accidentally overwrite this somewhere else. Also if we ever want to have this be dispatched from core we will need the prefix

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -72,14 +72,16 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as(relation, sql) -%}
+{% macro snowflake__create_view_as_with_temp_flag(temporary, relation, sql) -%}
```
dbeatty10, line 75: Part 1 of 3 of multi-part suggestion (completely untested!).

Basically, make the `temporary` field the final parameter and give it a default value.

```suggestion
{% macro snowflake__create_view_as_with_temp_flag(relation, sql, temporary=True) -%}
```

If this suggestion works, you could consider just using this as the new definition of `snowflake__create_view_as` and skip the independent `snowflake__create_view_as_with_temp_flag` macro altogether.

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -92,6 +94,10 @@
   );
 {% endmacro %}
 
+{% macro snowflake__create_view_as(relation, sql) -%}
+  {{ snowflake__create_view_as_with_temp_flag(False, relation, sql) }}
```
dbeatty10, line 98: Part 2 of 3 of multi-part suggestion:

```suggestion
  {{ snowflake__create_view_as_with_temp_flag(relation, sql, temporary=False) }}
```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -60,7 +60,7 @@
     {#-- Create the temp relation, either as a view or as a temp table --#}
     {% if tmp_relation_type == 'view' %}
         {%- call statement('create_tmp_relation') -%}
-          {{ create_view_as(tmp_relation, compiled_code) }}
+          {{ snowflake__create_view_as_with_temp_flag(True, tmp_relation, compiled_code) }}
```
dbeatty10, line 63: Part 3 of 3 of multi-part suggestion:

```suggestion
          {{ snowflake__create_view_as_with_temp_flag(tmp_relation, compiled_code) }}
```

dbeatty10 (commented):
A couple ideas to consider.

dbt/include/snowflake/macros/adapters.sql
dbeatty10: 🧠 very smart to confirm the order

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -72,14 +72,16 @@
 )
 {% endmacro %}
 
-{% macro snowflake__create_view_as(relation, sql) -%}
+{% macro snowflake__create_view_as_with_temp_flag(relation, sql, temporary=False) -%}
```
colin-rogers-dbt, line 75: Feel like I'm missing something if `temporary` arg defaults to `False` and we don't provide this value in `snowflake__create_view_as` won't the if condition on 83 never be satisfied? Should this be sourced from the config?

colin-rogers-dbt, line 75: nit: to disambiguate the macro arg and the value of "temporary" can we rename to `is_temporary`?

dbt/include/snowflake/macros/adapters.sql
dbeatty10: To tie off this convo, `snowflake__create_view_as` has been retained, so there shouldn't be any backwards compatibility issues if anyone overrides this in their projects.

A new macro named `snowflake__create_view_as_with_temp_flag` was introduced, and it is utilized by `snowflake__create_view_as`.

colin-rogers-dbt (approved):

VersusFacit:
Verified using my same test case that the rename has not altered the solution behavior. Merging

PR #389, opened 2023-01-11
[CT-1507] create config for incremental models to use temp tables or views to fix regression in perf

### LINKED ISSUE #323: [CT-1507] [Regression] v1.3: views instead of temp tables cause performance issue

### Current Behavior

[A change](https://github.com/dbt-labs/dbt-snowflake/pull/93) has been introduced in dbt 1.3.0 that uses a view instead of a temporary table for incremental models when the merge incremental strategy is used.
I noticed this change because one of my hourly job is now taking 30 minutes vs 3 minutes previously.
I know that using the `delete+insert` strategy [would work to get temp tables back](https://github.com/dbt-labs/dbt-snowflake/issues/306#issuecomment-1314573497) but we did some tests and the performance of our jobs is better when we use merge.

Would it be possible to go back to temporary table instead of views ? Or at least to add a parameter to be able to use temporary tables when the incremental strategy is a merge [here](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/include/snowflake/macros/materializations/incremental.sql#L13) ?

### Steps To Reproduce

1. Install dbt-core and dbt-snowflake v1.3
2. Launch an incremental model with the merge incremental_strategy
3. See that the query launched in Snowflake uses `create or replace view` vs `create or replace temporary table` in v<=1.2
4. Compare the query time of the merge between the view and the temporary table

### Relevant log output

_No response_

### Environment

```markdown
- OS: Irrelevant
- Python: 3.9
- dbt-core (working version): <=1.2
- dbt-snowflake (working version): 1.3
- dbt-core (regression version): <=1.2
- dbt-snowflake (regression version): 1.3
```

### Additional Context

_No response_

McKnight-42:
resolves #323 

### Description

create a model level config for incremental models to allow choice between using views or temp tables during merge strategy to help with performance issues in some cases.

co-author: @dbeatty10 

### Checklist


McKnight-42:
Want to flag this interesting suggestion @dbeatty10 shared with me here https://github.com/dbt-labs/dbt-snowflake/issues/323#issuecomment-1382015544 as part of the review discussion. @Fleid any thoughts on this?

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -10,10 +10,16 @@
        for faster overall incremental processing.
   #} */
 
-  {% if language == 'sql' and (strategy in ('default', 'append', 'merge') or (unique_key is none)) %}
-    {{ return('view') }}
-  {% else %}  {#--  play it safe -- #}
-    {{ return('table') }}
+  {% if language == "sql" and strategy in ("default", "merge") and merge_tmp_relation_type == "table" %}
```
colin-rogers-dbt, line 13: can we simplify and do the `if language == "sql"` condition once?

colin-rogers-dbt: something like this maybe? not tested or anything...
```suggestion
{% if language == "sql"%}
 {% if strategy in ("default", "merge")%}
  {% if merge_tmp_relation_type == "table" %}
    {{ return("table") }}
  {% else %}
    {{ return("view") }}
 {% elif strategy == "append" %}
    {{ return("view") }}
 {% elif unique_key is none %}
    {{ return("view") }}
{% else %}
   {{ return("table") }}
```

colin-rogers-dbt (commented):
suggested an alternate, maybe cleaner conditional setup but otherwise LGTM

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: @colin-rogers-dbt What do you think about alternative pseudo code for doing the `language` condition only once?

It's largely the same as what you proposed. The main difference is that it inverts the `sql` condition so that it doesn't need any nested conditionals.

```
if language != "sql":
    return "table"
else if strategy in ('default', 'merge') and merge_tmp_relation_type == "table":
    return "table"
else if strategy in ('default', 'merge') and merge_tmp_relation_type == "view":
    return "view"
else if strategy in ('default', 'merge', 'append'):
    return "view"
else if unique_key is none:
    return "view"
else:
    return "table"
```

colin-rogers-dbt: The != logic is definitely much cleaner

colin-rogers-dbt: if `merge_tmp_relation_type == "table"` would we ever return "view"?

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -10,10 +10,18 @@
        for faster overall incremental processing.
   #} */
 
-  {% if language == 'sql' and (strategy in ('default', 'append', 'merge') or (unique_key is none)) %}
-    {{ return('view') }}
-  {% else %}  {#--  play it safe -- #}
-    {{ return('table') }}
+  {% if language != "sql" %}
+    {{ return("table") }}
+  {% elif strategy in ('default', 'merge') and merge_tmp_relation_type == "table"%}
```
dbeatty10, line 15: Jinja formatting nit to add missing whitespace:
```suggestion
  {% elif strategy in ('default', 'merge') and merge_tmp_relation_type == "table" %}
```

dbt/include/snowflake/macros/materializations/incremental.sql
colin-rogers-dbt: In that case I think we can actually simplify this further:
```
if language != "sql":
    return "table"
else if strategy in ('append'):
    return "view"
else if merge_tmp_relation_type == "table":
    return "table"
else if merge_tmp_relation_type == "view":
    return "view"
else if unique_key is none:
    return "view"
else:
    return "table"
```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -10,10 +10,18 @@
        for faster overall incremental processing.
   #} */
```
dbeatty10, line 11: @McKnight-42 we should update the comments in L3-L11 to align with our changes.

Maybe something generally like the following? There's lots of different ways we could write the comments here. A key thing to communicate is _why_ different strategies use either a view or a temporary table. It would be nice if we make it explicit rather than having to do detective work, especially in the case where `unique_key` is none.

(I didn't bother to limit the line lengths below while we are iterating on the exact wording, but we'll want to add appropriate new lines in the final commit.)

```
  /* {#
       High-level principles:
       If we are running multiple statements (DELETE + INSERT),
       and we want to guarantee identical inputs to both statements, then
       we _must_ first save the model query results as a temporary table (which presumably comes with a performance cost).
       If we are running a single statement (MERGE or INSERT alone),
       we _may_ save the model query definition as a view instead,
       for (presumably) faster overall incremental processing.

       Low-level specifics:
       Languages other than SQL (like Python) will use a temporary table.
       With the `default` strategy of `merge`, the user may choose between a temporary table and view (defaulting to view).
       The `append` strategy can use a view because it will run a single INSERT statement.
       When the `unique_key` is none, then we can use a view because it will run a single INSERT statement.
       Otherwise, play it safe by using a temporary table.
  #} */
```

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: do you think that would make the fact that` merge_tmp_relation_type` only technically applies to the merge/default strategy as they are the same in snowflake?

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: Sounds great! will push up the change to the wording asap

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: @colin-rogers-dbt Love the way you are trying to make the logic as simple as possible! We just need to make sure to cover each of the moving pieces.

My understanding is that dbt-snowflake supports [three different incremental strategies](https://github.com/dbt-labs/dbt-snowflake/blob/4b7ee7372d87efc2b9c6929203fcfb41930b0b3e/dbt/adapters/snowflake/impl.py#L236-L237):
1. `merge` (which is the `default`)
1. `append`
1. `delete+insert`

Something to point out: there's the largely implicit case when `strategy == "delete+insert"` to consider. In the current version of the PR, it comes into play when all other strategies have already been considered _and_ `unique_key is none`. If your suggestion is implemented as-is, then the configuration of `merge_tmp_relation_type` would affect `merge` and `delete+insert`, but it wouldn't affect `append`.

Leaning into your idea, I suspect it would be like:
```
if language != "sql":
    return "table"
else if strategy in "append":
    return "view"
else if strategy == "delete+insert" and unique_key is none:
    return "view"
else if merge_tmp_relation_type == "table":
    return "table"
else if merge_tmp_relation_type == "view":
    return "view"
else:
    return "table"
```

Or a more future-proof version presuming some other strategy might be added at a later date:
```
if language != "sql":
    return "table"
else if strategy in "append":
    return "view"
else if strategy == "delete+insert" and unique_key is none:
    return "view"
else if strategy in ("default", "merge") and merge_tmp_relation_type == "table":
    return "table"
else if strategy in ("default", "merge") and merge_tmp_relation_type == "view":
    return "view"
else:
    return "table"
```

Side note: [here](https://github.com/dbt-labs/dbt-snowflake/issues/323#issuecomment-1382015544) is a proposal to make both the logic and the documentation as simple as possible. It would need buy-in from @Fleid. That proposal is non-blocking for this PR though.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -1,19 +1,34 @@
...
-    {{ return('table') }}
+  {% if language != "sql" %}
+    {{ return("table") }}
+  {% elif strategy in ('default', 'merge') and merge_tmp_relation_type == "table" %}
+    {{ return("table") }}
+  {% elif strategy in ("default", "merge") and merge_tmp_relation_type == "view" %}
+    {{ return("view") }}
+  {% elif strategy in ('default', 'merge', 'append') %}
+    {{ return("view") }}
+  {% elif unique_key is none %}
```
dbeatty10, line 28: The conversation [here](https://github.com/dbt-labs/dbt-snowflake/pull/389#discussion_r1072532245) makes me think we should make this update in order to future-proof this logic:
```suggestion
  {% elif strategy == 'delete+insert' and unique_key is none %}
```

Otherwise, if some new strategy is added that uses multiple statements in its implementation (like a delete step and an insert step), it might get handled incorrectly accidentally when unique_key is none.

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: Awesome @McKnight-42 !

Could you choose a maximum line length for the comments and then apply it? It looks like it was using a max line length of 70 previously. Using either [80](https://stackoverflow.com/questions/578059/studies-on-optimal-code-width) or [88](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) seems reasonable also.

colin-rogers-dbt (approved):

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: we can remove the assignment, but will mean that `tmp_relation_type` must always be assigned manually and change at least the syntax for `delete+insert` exception logic slightly. doing this should get the current tests passing which were failing based on the assignment for `view` being a default changing previous good behavior

PR #392, opened 2023-01-12
Mp/reduce ci daily run intervals

VersusFacit:
### Description

Need to tweak jobs one more time. You need main and integration on both combinations of jobs.


emmyoop (approved):
Good catch!

Issue #147, opened 2022-04-25
[CT-552] [Bug] Rows_affected showing as 1 in run_results.json and in UI


crystalro0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

User upgraded to v1.0 and is not seeing the same number of rows_affected in run logging. 

Seen in 0.18.1 logging:
```
00:00:01 | 1 of 1 OK created incremental model GEN.foobar [ [32mSUCCESS 130692981 [0m in 175.96s]
```

in v1.0 logging:
```
00:00:01 | 1 of 1 OK created incremental model GEN.foobar [ [32mSUCCESS 1 [0m in 188.32s]
```

Screenshot of user's snowflake history shows ~130M rows.

### Expected Behavior

The rows_affected number should be displaying the actual number (unless bound by limit).

### Steps To Reproduce

1. Using 0.18.1 and snowflake: create job and run.
2. Upgrade to 1.0.
3. Re-run job.

### Relevant log output

```shell
Please ping in slack for additional information if needed. ZD ticket and logging available.
```

### Environment

```markdown
- dbt: 1.0.4
```

### What database are you using dbt with?

snowflake

PR #401, opened 2023-01-17
ADAP-281 - Raise warning, not error, for isolated `begin;` or `commit;`

### LINKED ISSUE #388: [CT-1805] [Regression] Transaction behaviour change in dbt-snowflake `>=1.2` results in an error when macros call `run_query('begin;')`

### Current Behavior

A summary of the issue is available here https://github.com/dbt-labs/dbt-snowflake/issues/304

With our 1.2 release - we "fixed" a previous regression that did not show the number of affected rows - this caused users who had macros that contained singular transaction keywords like the following to error:

```jinja
{% do run_query('commit;') %}
```

dbt version 1 releases prior to `1.2` (i.e. `1.0`, `1.1`) had worked fine with the above calls (albeit with the wrong number of affected rows for incrementals) so we might want to do something about gracefully handling this "breaking" but net-improvement change.

dbt version `>=1.0,<1.2` had no errors when macros had `{% do run_query('commit;') %}` - we should continue to support this behaviour gracefully as long as we're in v1.

### Steps To Reproduce

1. Project setup.

```yml
# dbt_project.yml
name: my_dbt_project
profile: snowflake
config-version: 2
version: 1.0

models:
  my_dbt_project:
    +materialized: table
```

```sql
-- models/foo.sql
{{
    config(
        post_hook = "{{ my_macro() }}"
    )
}}
select 1 as id

-- macros/my_macro.sql
{% macro my_macro() %}
    {% do run_query('begin;') %}
    {% do run_query('create table if not exists development.dbt_jyeo.tester (dbt_version varchar, updated_at timestamp);') %}
    {% set query %}
       insert into development.dbt_jyeo.tester values ('{{ dbt_version }}', current_timestamp()); 
    {% endset %}
    {% do run_query(query) %}
    {% do run_query('commit;') %}
{% endmacro %}
```

2. Run with a few different versions.

```
$ dbt --version && dbt run
installed version: 1.0.8
   latest version: 1.3.2

Your version of dbt is out of date! You can find instructions for upgrading here:
https://docs.getdbt.com/docs/installation

Plugins:
  - snowflake: 1.0.1 - Update available!
  Your version of dbt-snowflake is out of date! You can find instructions for upgrading here:
  https://docs.getdbt.com/dbt-cli/install/overview
22:02:06  Running with dbt=1.0.8
22:02:07  Found 1 model, 0 tests, 0 snapshots, 0 analyses, 181 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
22:02:07  
22:02:15  Concurrency: 8 threads (target='dev')
22:02:15  
22:02:15  1 of 1 START table model dbt_jyeo.foo........................................... [RUN]
22:02:21  1 of 1 OK created table model dbt_jyeo.foo...................................... [SUCCESS 1 in 5.23s]
22:02:21  
22:02:21  Finished running 1 table model in 13.90s.
22:02:21  
22:02:21  Completed successfully
22:02:21  
22:02:21  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

image

```sh
$ dbt --version && dbt run
Core:
  - installed: 1.1.2
  - latest:    1.3.2 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - snowflake: 1.1.0 - Update available!

  At least one plugin is out of date or incompatible with dbt-core.
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation
22:03:52  Running with dbt=1.1.2
22:03:52  Found 1 model, 0 tests, 0 snapshots, 0 analyses, 183 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
22:03:52  
22:03:58  Concurrency: 8 threads (target='dev')
22:03:58  
22:03:58  1 of 1 START table model dbt_jyeo.foo .......................................... [RUN]
22:04:03  1 of 1 OK created table model dbt_jyeo.foo ..................................... [SUCCESS 1 in 5.21s]
22:04:03  
22:04:03  Finished running 1 table model in 10.90s.
22:04:03  
22:04:03  Completed successfully
22:04:03  
22:04:03  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

image

```sh
Core:
  - installed: 1.2.4
  - latest:    1.3.2 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - snowflake: 1.2.0 - Update available!

  At least one plugin is out of date or incompatible with dbt-core.
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation
22:04:49  Running with dbt=1.2.4
22:04:49  Found 1 model, 0 tests, 0 snapshots, 0 analyses, 269 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
22:04:49  
22:04:55  Concurrency: 8 threads (target='dev')
22:04:55  
22:04:55  1 of 1 START table model dbt_jyeo.foo .......................................... [RUN]
22:04:59  1 of 1 ERROR creating table model dbt_jyeo.foo ................................. [ERROR in 3.84s]
22:04:59  
22:04:59  Finished running 1 table model in 0 hours 0 minutes and 9.30 seconds (9.30s).
22:04:59  
22:04:59  Completed with 1 error and 0 warnings:
22:04:59  
22:04:59  Runtime Error in model foo (models/foo.sql)
22:04:59    Tried to run an empty query on model 'model.my_dbt_project.foo'. If you are conditionally running
22:04:59    sql, eg. in a model hook, make sure your `else` clause contains valid sql!
22:04:59    
22:04:59    Provided SQL:
22:04:59    
22:04:59        begin;
22:04:59      
22:04:59  
22:04:59  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.8.2
- dbt-core (working version): <1.2
- dbt-snowflake (working version): <1.2
- dbt-core (regression version): >=1.2
- dbt-snowflake (regression version): >=1.2
```

### Additional Context

_No response_

jtcohen6:
resolves #388

**Goal:** Provide a subset of `dbt-snowflake` users with a much nicer upgrade path to v1.2+. If we're happy with the change, and think it's sufficiently precise, we should backport to v1.2-v1.4.

### Description

Following @jeremyyeo's repro case, this will now succeed without errors, but it will raise a warning:
```
10:58:29  Found 1 model, 0 tests, 0 snapshots, 0 analyses, 305 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
10:58:29
10:58:33  Concurrency: 4 threads (target='default')
10:58:33
10:58:33  1 of 1 START sql table model test16739531067326963395_test_isolated_begin_commit.my_model  [RUN]
10:58:35  Snowflake adapter: [WARNING]: Explicit transactional logic should be used only to wrap DML
logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should be
placed directly before and after your DML statement, rather than in separate
statement calls or run_query() macros.
10:58:37  Snowflake adapter: [WARNING]: Explicit transactional logic should be used only to wrap DML
logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should be
placed directly before and after your DML statement, rather than in separate
statement calls or run_query() macros.
10:58:38  1 of 1 OK created sql table model test16739531067326963395_test_isolated_begin_commit.my_model  [SUCCESS 1 in 4.92s]
10:58:38
10:58:38  Finished running 1 table model in 0 hours 0 minutes and 9.15 seconds (9.15s).
10:58:38
10:58:38  Completed successfully
```

#### Questions
- The warning will appear twice, once for `BEGIN;` and once for `COMMIT;`. Should we raise it only for `BEGIN;`? Given how niche this behavior is, I don't think we need to invest a ton of energy in making it pretty.
- Where should the functional test for this go? I just stuck it at the top-level `tests/functional` for now

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/connections.py
+++ b/dbt/adapters/snowflake/connections.py
@@ -438,6 +438,16 @@ class SnowflakeConnectionManager(SQLConnectionManager):
             table = dbt.clients.agate_helper.empty_table()
         return response, table
 
+    def add_standard_query(self, sql, auto_begin, bindings, abridge_sql_log):
+        # This is the happy path for a single query. Snowflake has a few odd behaviors that
+        # require preprocessing within the 'add_query' method below.
+        return super().add_query(
+            self._add_query_comment(sql),
+            auto_begin=auto_begin,
+            bindings=bindings,
+            abridge_sql_log=abridge_sql_log,
+        )
+
     def add_query(self, sql, auto_begin=True, bindings=None, abridge_sql_log=False):
         connection = None
         cursor = None
@@ -462,38 +472,36 @@ class SnowflakeConnectionManager(SQLConnectionManager):
             if without_comments != "":
                 stripped_queries.append(query)
 
+        if set(query.lower() for query in stripped_queries).issubset({"begin;", "commit;"}):
+            # if all we get is `begin;` and/or `commit;`
+            # raise a warning, then run as standard queries to avoid an error downstream
+            message = (
+                "Explicit transactional logic should be used only to wrap "
+                "DML logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should "
+                "be placed directly before and after your DML statement, rather than in separate "
+                "statement calls or run_query() macros."
+            )
+            logger.warning(line_wrap_message(warning_tag(message)))
+            for query in stripped_queries:
+                connection, cursor = self.add_standard_query(
+                    query, auto_begin, bindings, abridge_sql_log
+                )
+
         for individual_query in stripped_queries:
             # Even though we turn off transactions by default for Snowflake,
             # the user/macro has passed them *explicitly*, probably to wrap a DML statement
-            # Assuming this BEGIN/COMMIT is not all alone - let their wish be granted!
             # This also has the effect of ignoring "commit" in the RunResult for this model
             # https://github.com/dbt-labs/dbt-snowflake/issues/147
-            # (Otherwise, if this 'BEGIN' or 'COMMIT' query *is* all by itself, let it be run as a standalone query)
-            if len(stripped_queries) == 1 and individual_query.lower() in ("begin;", "commit;"):
-                message = (
-                    "Explicit transactional logic should be used only to wrap "
-                    "DML logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should "
-                    "be placed directly before and after your DML statement, rather than in separate "
-                    "statement calls or run_query() macros."
-                )
-                logger.warning(line_wrap_message(warning_tag(message)))
+            if individual_query.lower() == "begin;":
+                super().add_begin_query()
+            elif individual_query.lower() == "commit;":
+                super().add_commit_query()
             else:
-                if individual_query.lower() == "begin;":
-                    super().add_begin_query()
-                    continue
-
-                elif individual_query.lower() == "commit;":
-                    super().add_commit_query()
-                    continue
-
-            # add a query comment to *every* statement
-            # needed for models with multi-step materializations
-            # https://github.com/dbt-labs/dbt-snowflake/issues/140
-            individual_query = self._add_query_comment(individual_query)
-
-            connection, cursor = super().add_query(
-                individual_query, auto_begin, bindings=bindings, abridge_sql_log=abridge_sql_log
-            )
+                # This adds a query comment to *every* statement
+                # https://github.com/dbt-labs/dbt-snowflake/issues/140
+                connection, cursor = self.add_standard_query(
+                    individual_query, auto_begin, bindings, abridge_sql_log
+                )
 
         if cursor is None:
             conn = self.get_thread_connection()
--- a/dbt/adapters/snowflake/connections.py
+++ b/dbt/adapters/snowflake/connections.py
@@ -42,6 +42,7 @@ from dbt.adapters.sql import SQLConnectionManager  # type: ignore
 from dbt.events import AdapterLogger  # type: ignore
 from dbt.events.functions import warn_or_error
 from dbt.events.types import AdapterEventWarning
+from dbt.ui import line_wrap_message, warning_tag
 
 
 logger = AdapterLogger("Snowflake")
@@ -447,32 +448,43 @@ class SnowflakeConnectionManager(SQLConnectionManager):
             bindings = tuple(bindings)
 
         queries = self._split_queries(sql)
-
-        for individual_query in queries:
+        stripped_queries = []
+        for query in queries:
             # hack -- after the last ';', remove comments and don't run
             # empty queries. this avoids using exceptions as flow control,
             # and also allows us to return the status of the last cursor
             without_comments = re.sub(
                 re.compile(r"(\".*?\"|\'.*?\')|(/\*.*?\*/|--[^\r\n]*$)", re.MULTILINE),
                 "",
-                individual_query,
+                query,
             ).strip()
 
-            if without_comments == "":
-                continue
+            if without_comments != "":
+                stripped_queries.append(without_comments)
 
+        for individual_query in stripped_queries:
             # Even though we turn off transactions by default for Snowflake,
             # the user/macro has passed them *explicitly*, probably to wrap a DML statement
-            # Let their wish be granted!
+            # Assuming this BEGIN/COMMIT is not all alone - let their wish be granted!
             # This also has the effect of ignoring "commit" in the RunResult for this model
             # https://github.com/dbt-labs/dbt-snowflake/issues/147
-            if individual_query.lower() == "begin;":
-                super().add_begin_query()
-                continue
+            # (Otherwise, if this 'BEGIN' or 'COMMIT' query *is* all by itself, let it be run as a standalone query)
+            if len(stripped_queries) == 1 and individual_query.lower() in ("begin;", "commit;"):
+                message = (
+                    "Explicit transactional logic should be used only to wrap "
+                    "DML logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should "
+                    "be placed directly before and after your DML statement, rather than in separate "
+                    "statement calls or run_query() macros."
+                )
+                logger.warning(line_wrap_message(warning_tag(message)))
+            else:
+                if individual_query.lower() == "begin;":
+                    super().add_begin_query()
+                    continue
 
-            elif individual_query.lower() == "commit;":
-                super().add_commit_query()
-                continue
+                elif individual_query.lower() == "commit;":
+                    super().add_commit_query()
+                    continue
 
             # add a query comment to *every* statement
             # needed for models with multi-step materializations
--- a/dbt/adapters/snowflake/connections.py
+++ b/dbt/adapters/snowflake/connections.py
@@ -491,9 +491,7 @@ class SnowflakeConnectionManager(SQLConnectionManager):
             )
             return re.sub(without_comments_re, "", query).strip()
 
-        return [
-            strip_query(query) for query in self._split_queries(sql) if strip_query(query) != ""
-        ]
+        return [query for query in self._split_queries(sql) if strip_query(query) != ""]
 
     def _add_begin_commit_only_queries(
         self, queries: List[str], **kwargs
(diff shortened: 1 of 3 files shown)


dbt/adapters/snowflake/connections.py
```diff
@@ -435,32 +436,43 @@ def add_query(self, sql, auto_begin=True, bindings=None, abridge_sql_log=False):
...
             # the user/macro has passed them *explicitly*, probably to wrap a DML statement
-            # Let their wish be granted!
+            # Assuming this BEGIN/COMMIT is not all alone - let their wish be granted!
             # This also has the effect of ignoring "commit" in the RunResult for this model
             # https://github.com/dbt-labs/dbt-snowflake/issues/147
-            if individual_query.lower() == "begin;":
-                super().add_begin_query()
-                continue
+            # (Otherwise, if this 'BEGIN' or 'COMMIT' query *is* all by itself, let it be run as a standalone query)
+            if len(stripped_queries) == 1 and individual_query.lower() in ("begin;", "commit;"):
```
mikealfare, line 460: I think this check should happen outside of the for loop. Basically "if you handed me a single commit; or begin;, then warn. otherwise start processing some queries".

jtcohen6: @mikealfare There's some trickiness here. If we're handed a single query AND that query is `begin;` or `commit;`, we want to pass it through, to avoid this error:
```
[DEBUG] stdout_log: 16:55:24.961406 [debug] [Thread-1  ]: Runtime Error in model my_model (models/my_model.sql)
  Tried to run an empty query on model 'model.test.my_model'. If you are conditionally running
  sql, eg. in a model hook, make sure your `else` clause contains valid sql!

  Provided SQL:

      begin;
```
But if we're handed _multiple_ queries (semicolon-delimited), which are wrapped in a `begin;` and a `commit;`, then we want to do the handling below, that converts them into `super().add_begin_query()` and `super().add_commit_query()`. (This was the fix for https://github.com/dbt-labs/dbt-snowflake/issues/147).

So it is possible to pull this warning out of the `for` loop, but we still need an added conditional check within the `for` loop. Do you prefer it being written this way?
```python
        # Even though we turn off transactions by default for Snowflake,
        # the user/macro has passed them *explicitly*, probably to wrap a DML statement.
        # Assuming this BEGIN/COMMIT is not all alone - let their wish be granted!
        # This also has the effect of ignoring "commit" in the RunResult for this model
        # https://github.com/dbt-labs/dbt-snowflake/issues/147
        
        # (However, if this 'BEGIN' or 'COMMIT' query *is* all by itself, raise a warning and let it be run as a standalone query)
        if len(stripped_queries) == 1 and stripped_queries[0].lower() in ("begin;", "commit;"):
            message = (
                "Explicit transactional logic should be used only to wrap "
                "DML logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should "
                "be placed directly before and after your DML statement, rather than in separate "
                "statement calls or run_query() macros."
            )
            logger.warning(line_wrap_message(warning_tag(message)))

        for individual_query in stripped_queries:
            if len(stripped_queries) > 1 and individual_query.lower() == "begin;":
                super().add_begin_query()
                continue

            elif len(stripped_queries) > 1 and individual_query.lower() == "commit;":
                super().add_commit_query()
                continue
```

mikealfare: I see what you're saying. So then you'd need to put the rest in an `else` and add a step to the "exception" condition. That would also get rid of the need for `continue`, which improves readability in my opinion. Basically I'd rather separate the "exception" state (a single `begin;`/`commit;`) from the "happy path"/"normal" state. So it could look like this:
```
for query in queries:
    ...

def add_standard_query(query_def):
    return super().add_query(
        self._add_query_comment(query_def),
        auto_begin,
        bindings=bindings,
        abridge_sql_log=abridge_sql_log
    )

if len(stripped_queries) == 1 and stripped_queries[0].lower() in ["begin;", "commit;"]:
    # comments speaking to exception case
    message = (...)
    logger.warning(line_wrap_message(warning_tag(message)))
    connection, cursor = add_standard_query(stripped_queries[0])
else:
    for query in queries:
        # comments, if any, speaking to happy path
        if query.lower() == "begin;":
            super().add_begin_query()
        elif query.lower() == "commit;":
            super().add_commit_query()
        else:
            connection, cursor = add_standard_query(query)

if cursor is None:
    ...

return connection, cursor
```

It appears that the last exception logic is only triggered if we receive a series of `begin;` or `commit;` queries. So our three scenarios are:

1. Single `begin;` or `commit;` > let it pass through as a standard query with no error
2. A series of queries which are all either `begin;` or `commit;` > throw a DbtRuntimeError
3. A series of queries, at least one of which is not `begin;` or `commit;` > the cursor is set

If that's the case, should we combine the first two scenarios? It could look like this:
```
for query in queries:
    ...

def add_standard_query(query_def):
    return super().add_query(
        self._add_query_comment(query_def),
        auto_begin,
        bindings=bindings,
        abridge_sql_log=abridge_sql_log
    )

if set(query.lower() for query in stripped_queries).issubset(set("begin;", "commit;")):
    # if all we get are a series of `begin;` and/or `commit;`, return one of them as a standard query to avoid an error downstream
    # comments speaking to exception case
    message = (...)
    logger.warning(line_wrap_message(warning_tag(message)))
    return add_standard_query(stripped_queries[0])  # do we even want to return a query?

for query in queries:
    # comments, if any, speaking to happy path
    if query.lower() == "begin;":
        super().add_begin_query()
    elif query.lower() == "commit;":
        super().add_commit_query()
    else:
        connection, cursor = add_standard_query(query)

return connection, cursor
```

jtcohen6: @mikealfare I like it! Much cleaner separation of the logic. I just pushed up a change in this vein, let me know what you think

dbt/adapters/snowflake/connections.py
```diff
@@ -437,6 +438,16 @@ def execute(
             table = dbt.clients.agate_helper.empty_table()
         return response, table
 
+    def add_standard_query(self, sql, auto_begin, bindings, abridge_sql_log):
```
mikealfare, line 441: We just added this to `dbt-core` for a similar reason:
https://github.com/dbt-labs/dbt-core/blob/24ca76ea58210c3bec367734dfebfc974b58be21/core/dbt/adapters/sql/connections.py#L158

It might be worth updating the definition in `dbt-core` with defaults that match `add_query()` and then referencing that here.

dbt/adapters/snowflake/connections.py
```diff
@@ -447,41 +458,50 @@ def add_query(self, sql, auto_begin=True, bindings=None, abridge_sql_log=False):
...
+                "Explicit transactional logic should be used only to wrap "
+                "DML logic (MERGE, DELETE, UPDATE, etc). The keywords BEGIN; and COMMIT; should "
+                "be placed directly before and after your DML statement, rather than in separate "
+                "statement calls or run_query() macros."
+            )
+            logger.warning(line_wrap_message(warning_tag(message)))
+            for query in stripped_queries:
+                connection, cursor = self.add_standard_query(
+                    query, auto_begin, bindings, abridge_sql_log
+                )
```
mikealfare, line 488: I think we need a `return` here, right? If we get a set queries that are only `begin;` and/or `commit;`, then we want the this condition and that's all. In other words, we don't want the for loop below to run and execute `super().add_begin_query()` and `super().add_commit_query()`. I might not be remembering the requirements correctly.

dbt/adapters/snowflake/connections.py
mikealfare: Two items, seen below. I think there's a missing `return`, and we implemented something similar to the `add_standard_query()` method in `dbt-core`.

dbt/adapters/snowflake/connections.py
mikealfare: I just re-read the ticket. It looks like we want to backport this into `1.4.latest`, `1.3.latest`, and `1.2.latest`. Is that right? If so, we can't use the new method since that is only in `v1.5`. In that case, I'm fine with this.

mikealfare (approved):

PR #403, opened 2023-01-18
[CT-1507] part 2: Expand configuration to all strategies

### LINKED ISSUE #402: [CT-1843] [Feature] Generic `tmp_relation_type` config for incremental models

### Describe the feature

https://github.com/dbt-labs/dbt-snowflake/pull/389 was merged, which adds the `merge_tmp_relation_type` configuration that only applies to the `merge` strategy.

So we'll probably want to make a decision on the following proposal (from [here](https://github.com/dbt-labs/dbt-snowflake/issues/323#issuecomment-1382015544)) before the RC2 that includes that feature is released.

---

As @McKnight-42 and I iterated on the implementation of https://github.com/dbt-labs/dbt-snowflake/issues/323, a question came up how we should handle the `merge_tmp_relation_type` config for strategies other than the `default` strategy of `merge`.

Consider the `append` strategy:
- How _should_ it behave if the user specifies `merge_tmp_relation_type=table` and `strategy=append`?
    - we currently would use a temporary `view`

### Three options
We have three main options:
1. Raise an exception that the `merge_tmp_relation_type` configuration doesn't apply to this strategy
1. (Silently?) ignore the `merge_tmp_relation_type` configuration 
1. Expand the scope so that a `incremental_tmp_relation_type` config also applies to the rest of the incremental strategies (but only when provided)

### Trade-offs
Primary trade-off of each option (in corresponding order):
1. Could be frustrating to the user if they can control the type of temporary object for only the `merge` strategy but not the others.
1. Could be confusing if they specify `table` but it uses a `view` anyways.
1. A user could override to use a temporary object in a way that is less efficient than the alternative. (i.e., configuring `table` in a case where `view` would have been faster.) Also, the discerning eye will see that that `language!=sql` will still use `table` no matter what the user says.

### Proposal
So I'd like to propose that we actually do the third option! This would expand the scope slightly to cover other incremental strategies also.

One advantage is that writing the documentation for the third option would be the most simple. It would also be easiest to understand from a user perspective:
- when this option is provided (for SQL models), then the adapter will do whatever you say
- otherwise, then the adapter will do whatever it thinks best

### Trying on Option Three

Suppose the new config name is `tmp_relation_type`. Then the pseudo logic would be:

```
if language == "sql" and tmp_relation_type == "table":
    return "table"
else if language == "sql" and tmp_relation_type == "view":
    return "view"
else:
    return whatever the current logic returns
```

> As a user, I prefer the name `tmp_relation_type` over `incremental_tmp_relation_type` since it is shorter and not redundant with the materialization type of `incremental`)

Expanding this out to include all the current logic (and simplifying the `language` a teensy bit):
```
if language == "python":
    return "table"
else if tmp_relation_type == "table":
    return "table"
else if tmp_relation_type == "view":
    return "view"
else if strategy in ('default', 'merge', 'append'):
    return "view"
else if unique_key is none:
    return "view"
else:
    return "table"
```

> Note that `delete+insert` is the only strategy not explicitly referenced by name and it will use `view` unless a `unique_key` is provided.

### Current Logic in PR [#389](https://github.com/dbt-labs/dbt-snowflake/pull/389)

If we keep the original decision to have a `merge_tmp_relation_type` config that only applies to the `merge` strategy, then the logic could be essentially like:
```
# TODO:
# Either raise an error if `merge_tmp_relation_type` is configured for a strategy other than `default` or `merge`
# or ignore altogether

if language == "python":
    return "table"
else if strategy in ('default', 'merge') and merge_tmp_relation_type == "table":
    return "table"
else if strategy in ('default', 'merge') and merge_tmp_relation_type == "view":
    return "view"
else if strategy in ('default', 'merge', 'append'):
    return "view"
else if unique_key is none:
    return "view"
else:
    return "table"
```

Ultimately, the two variations aren't that substantially different from each other. However, if we do ever want to come back and allow strategies other than `merge` to configure their temp object type, then the implementation in code will get uglier, and the documentation will be more complicated.

If we allow all incremental strategies to provide this config in one fell swoop right now, then we'll complete the offering of a  balance of rational defaults along with configurability for intermediate temp objects used by incremental models.

See above.

The main alternative is a `merge_tmp_relation_type` config that is meant to only ever apply to the `merge` strategy and never apply to any other incremental strategy.

### Who will this benefit?

If we ever foresee being able to configure the temporary object type (table vs. view) for an incremental strategy other than `merge`, it would be better to do this now rather than later.

Happy to help however

### Anything else?

_No response_

McKnight-42:
resolves #402 

### Description
Take the new logic from #389 and expand and configuration to all incremental strategies to allow simplification of conditional logic for readability.

co-author: @dbeatty10 

### Checklist


dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -20,15 +20,15 @@
...
-  {% elif strategy in ("default", "merge") and merge_tmp_relation_type == "table" %}
+  {% elif tmp_relation_type == "table" %}
     {{ return("table") }}
-  {% elif strategy in ("default", "merge") and merge_tmp_relation_type == "view" %}
+  {% elif tmp_relation_type == "view" %}
     {{ return("view") }}
   {% elif strategy in ("default", "merge", "append") %}
     {{ return("view") }}
-  {% elif strategy == "delete+insert" and unique_key is none %}
+  {% elif unique_key is none %}
```
dbeatty10, line 31: Could we change the minimal number of lines to preserve as much of the logic we worked on as possible?

Something like:
```sql
  {% if language != "sql" %}
    {{ return("table") }}
  {% elif tmp_relation_type == "table" %}
    {{ return("table") }}
  {% elif tmp_relation_type == "view" %}
    {{ return("view") }}
  {% elif strategy in ("default", "merge", "append") %}
    {{ return("view") }}
  {% elif strategy == "delete+insert" and unique_key is none %}
```

McKnight-42: sure can convert back some, will wind up with a few extra lines through to incorporate the conditions for the exceptions we wish to raise

dbeatty10: The exceptions can probably be in a separate block prior to this.

Something like:
```sql
# Handle un-happy path first
if language == 'python' and incremental_tmp_relation_type == 'view':
     raise exception
elif incremental_strategy == 'delete+insert' and incremental_tmp_relation_type == 'table':
     raise exception

# Handle happy path second
if language != "sql":
    return "table"
elif ...
```

McKnight-42: sounds good, will add it now, have we decided on the config name? `tmp_relation_type` or `incremental_tmp_relation_type` btw?

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -20,11 +20,19 @@
        Otherwise, play it safe by using a temporary table.
   #} */
 
+  {% if language == "python" and tmp_relation_type == "table" %}
```
dbeatty10, line 23: Is this supposed to be the opposite?
```suggestion
  {% if language == "python" and tmp_relation_type != "table" %}
```

McKnight-42: depends on the use case honestly, if just doing a insert then a temp view would be quicker, but if you want to do a actual delete and insert it must be a temp table. 

curious if this truly needs to be raised or if we can somehow make the use cases beyond easy to understand

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -20,11 +20,19 @@
...
+    {% do exceptions.raise_compiler_error("Python models currently only support temp tables. ") %}
+  {% endif %}
+
+  {% if strategy == "delete+insert" and tmp_relation_type == "table" %}
+    {% do exceptions.raise_compiler_error("Possiblity of inconsistant results when using temp table with delete+insert") %}
```
dbeatty10, line 28: Could we explicitly tell the user what they configured versus what the allowed value is? We'll also want to fix the spellings of "possibility" and "inconsistent" (if we keep both of those in the final message).

Something like the following (which I didn't test out)?

```suggestion
    {% do exceptions.raise_compiler_error(f"Python models only support 'table' for 'tmp_relation_type', but '{tmp_relation_type}' was specified. ") %}
  {% endif %}

  {% if strategy == "delete+insert" and unique_key is not none and tmp_relation_type != "table" %}
    {% do exceptions.raise_compiler_error(f"In order to guarantee consistent results when 'unique_key' is given, the 'delete+insert' strategy only supports 'table' for 'tmp_relation_type'  but '{tmp_relation_type}' was specified.") %}
```

Alternatively, we can say something very simple like "'tmp_relation_type' is not allowed for Python models".

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -20,11 +23,27 @@
        Otherwise, play it safe by using a temporary table.
   #} */
 
+  {% if language == "python" and tmp_relation_type != "None" %}
```
dbeatty10, line 26: I didn't check, but should this be:
```suggestion
  {% if language == "python" and tmp_relation_type is not none %}
```

McKnight-42: very likley swapping between jinja language and everything else is fun haha. change made

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: updated messaging, did some local testing will. continue in a bit looking good thus far

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -1,5 +1,8 @@
 {% macro dbt_snowflake_get_tmp_relation_type(strategy, unique_key, language) %}
-  {%- set merge_tmp_relation_type = config.get('merge_tmp_relation_type', default="view") -%}
+  {% if language == "sql"%}
+    {%- set tmp_relation_type = config.get('tmp_relation_type', default="view") -%}
```
dbeatty10, line 3: Is this necessary? Or can we just delete this line altogether?

In particular, I think this line will cause problems when:
```
strategy == "delete+insert" and tmp_relation_type != "table" and unique_key is not none
```

My assumption is that we can specify the subsequent logic in a way that will handle all the relevant cases if `tmp_relation_type` is not provided.

```suggestion
```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -20,11 +21,27 @@
        Otherwise, play it safe by using a temporary table.
```
dbeatty10, line 21: @McKnight-42 could you update the docs in this section?

Maybe add a line here like:
```
       Low-level specifics:
       If an invalid option is specified, then raise an error with an applicable explanation
```

And another update like:
```
       When the unique_key is none, the delete+insert strategy
       can use a view because it will run a single INSERT statement.
```

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: I think that we can write the logic in such a way as to allow `tmp_relation_type` to be completely optional to the user.

In fact, we _want_ it to be completely optional rather than required.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -30,7 +28,7 @@
     ) %}
   {% endif %}
 
-  {% if strategy == "delete+insert" and tmp_relation_type != "table" and unique_key is not none %}
+  {% if strategy == "delete+insert" and tmp_relation_type == "view" and unique_key is not none %}
```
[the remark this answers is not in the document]
McKnight-42: This was causing issues with some existing testing it looked like so playing around with changes

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -11,20 +10,41 @@
...
        Languages other than SQL (like Python) will use a temporary table.
        With the default strategy of merge, the user may choose between a temporary
        table and view (defaulting to view).
-       The append strategy can use a view because it will run a single INSERT statement.
-       When the unique_key is none,
-       then we can use a view because it will run a single INSERT statement.
-       Otherwise, play it safe by using a temporary table.
+
+       The append strategy can use a view because it will run a single INSERT statement
+       when the unique_key is none.
```
dbeatty10, line 21: It looks like "when the unique_key is none" got accidentally duplicated and applied to the `append` strategy. I think it was only intended to apply to the `delete+insert` strategy.
```suggestion
```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -11,20 +10,41 @@
...
+       If an invalid option is specified, then we will raise an
+       excpetion with corresponding message.
+
        Languages other than SQL (like Python) will use a temporary table.
        With the default strategy of merge, the user may choose between a temporary
        table and view (defaulting to view).
-       The append strategy can use a view because it will run a single INSERT statement.
-       When the unique_key is none,
-       then we can use a view because it will run a single INSERT statement.
-       Otherwise, play it safe by using a temporary table.
```
dbeatty10, line 20: Why was the "play it safe comment removed?

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: the append strategy part actually was already there I believe can remove it though if it seems unnecessary.

dbt/include/snowflake/macros/materializations/incremental.sql
McKnight-42: in the high level principles it looks as though we go into detail as to why we would use the temporary table in regards to the `delete+insert` option.

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: Yeah, I think we'll want to remove it because it's actually inaccurate.

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: I'm not against completely re-organizing it per se. But I also don't see a need to re-org it dramatically. My default approach is to change as little as necessary.

I'd vote for adding this back:
```
 Otherwise, play it safe by using a temporary table.
 ```

**Review by dbeatty10 — CHANGES_REQUESTED** [review]
@McKnight-42 do you know why some CI tests are failing?

File: dbt/include/snowflake/macros/materializations/incremental.sql [hunk 10]
```diff
@@ -18,11 +18,9 @@
        With the default strategy of merge, the user may choose between a temporary
        table and view (defaulting to view).
 
-       The append strategy can use a view because it will run a single INSERT statement
-       when the unique_key is none.
-
```
dbeatty10: @McKnight-42 I think we deleted a bit too much here.

We should keep this:
```
       The append strategy can use a view because it will run a single INSERT statement
```

But make sure this doesn't sneak back in:
```
       when the unique_key is none.
```

McKnight-42:
@dbeatty10 tests are failing because as of right now the default value being passed in for the condtional logic is none which fails for the `delete+insert` check due to not being a `table` by configuration. playing around with it right now.

could probably get passing by just specifying == "view" but as mentioned above doesn't handle other possible unsupported configs later on

could also get passing by adding the configuration to models being tested (looking for which ones they are specifically already incase that is what we want to do)

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -11,20 +11,39 @@
...
+       If an invalid option is specified, then we will raise an
+       excpetion with corresponding message.
+
        Languages other than SQL (like Python) will use a temporary table.
        With the default strategy of merge, the user may choose between a temporary
        table and view (defaulting to view).
-       The append strategy can use a view because it will run a single INSERT statement.
-       When the unique_key is none,
-       then we can use a view because it will run a single INSERT statement.
+
```
dbeatty10, line 20: We should add this back in:
```suggestion
       The append strategy can use a view because it will run a single INSERT statement.

```

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -11,20 +11,41 @@
...
   #} */
 
+  {% if language == "python" and tmp_relation_type is not none %}
+    {% do exceptions.raise_compiler_error(
+      "Python models currently only support 'table' for tmp_relation_type but "
+       ~ tmp_relation_type ~ " was specified."
+    ) %}
+  {% endif %}
+
+  {% if strategy == "delete+insert" and tmp_relation_type != "table" and unique_key is not none %}
```
dbeatty10, line 35: The CI output [here](https://github.com/dbt-labs/dbt-snowflake/actions/runs/3952937214/jobs/6768632525) gives us some hints of what we'd need to change:
```
21:10:02  Completed with 1 error and 0 warnings:
21:10:02  
21:10:02  Compilation Error in model model (models-unquoted/model.sql)
21:10:02    In order to maintain consistent results when `unique_key` is not none,
21:10:02          the `delete+insert` strategy only supports `table` for `tmp_relation_type` but None was specified.
```

Although a bit hard to read, using the [Jinja `default` filter](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.default) might work:
```suggestion
  {% if strategy == "delete+insert" and (tmp_relation_type|default("table") != "table" and unique_key is not none %}
```

A different way to write the same thing:
```suggestion
  {% if strategy == "delete+insert" and tmp_relation_type is not none and tmp_relation_type != "table" and unique_key is not none %}
```

McKnight-42: like minds, had the second implementation on my list of things to try, currently trying moving the exceptions after our actual choice logic that way it's testing after we choose rather than before just want to make sure the main condtional logic works the way we think as well.

McKnight-42: some quick testing makes it seem like `None` counts as a valid value and so wouldn't swap out to the default in the filter method

testing second method, now didn't like how the moving of the exceptions acted.

dbt/include/snowflake/macros/materializations/incremental.sql
dbeatty10: My preference is `tmp_relation_type`, and I didn't see any arguments to the contrary [here](https://github.com/dbt-labs/dbt-snowflake/issues/402#issuecomment-1387300212).

colin-rogers-dbt (approved):
this looks solid and much more readable

tests/integration/simple_copy_test/test_simple_copy.py
```diff
@@ -266,7 +266,8 @@ def test__snowflake__incremental_overwrite(self):
         # Setting the incremental_strategy should make this succeed
         self.use_default_project({
             "models": {
-                "incremental_strategy": "delete+insert"
+                "incremental_strategy": "delete+insert",
+                "tmp_relation_type": "table"
```
dbeatty10, line 270: Is this change necessary? Or was it just temporary and can be restored back to the original now?
```suggestion
                "incremental_strategy": "delete+insert"
```

.changes/unreleased/Under the Hood-20230119-103909.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Under the Hood
+body: 'expand logic from #389 to try and make to config option tmp_relation_type as
+  optional as possible while maintining saftey of table as default and adding a few
+  exceptions to raise in rare casses with clear messages on fix.'
```
dbeatty10, line 4: Can we make this more clear and succinct?

One suggestion
```suggestion
body: 'Enable a `tmp_relation_type` config for applicable incremental models'
```

dbeatty10 (changes requested):
The logic _looks_ good  to me and also appears like it is working.

Two suggestions included below:
1. Let's make the changelog entry clear and succinct
2. Can we remove the test customization and confirm it passes CI?

tests/integration/simple_copy_test/test_simple_copy.py
McKnight-42: removed config testing locally and will push up

tests/integration/simple_copy_test/test_simple_copy.py
```diff
@@ -266,7 +266,7 @@ def test__snowflake__incremental_overwrite(self):
         # Setting the incremental_strategy should make this succeed
         self.use_default_project({
             "models": {
-                "incremental_strategy": "delete+insert"
+                "incremental_strategy": "delete+insert",
```
dbeatty10, line 269: I know this is the nit of all nits, but can we remove this extraneous comma to restore the original configuration?
```suggestion
                "incremental_strategy": "delete+insert"
```

dbeatty10 (approved):
All CI is passing and the logic looks good!

Is there any chance we can remove that extra comma before merging though? 🙏 Suggestion below. All that would be needed is to click the "Commit suggestion" button!

PR #453, opened 2023-02-07
remove graph_selection test

McKnight-42:
resolves #

### Description

remove duplicated test
### Checklist


McKnight-42:
There is a core version of this tests https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/graph_selection/test_graph_selection.py up all setup in general is the same but some of the running methods would be different between the two because of differences in how model names are represented, would we still like to pull in the `SelectionFixtures` for this and move test in core to adapter zone or is that not worth it since core also has some other tests in that section

VersusFacit:
It's only worth it if the adapter zone test would be used in this test. But, I don't think that's possible because Snowflake uses it's own settings right? If I misheard you earlier, then yes, we should move the core to the adapter zone and reimport some of those test classes here.

McKnight-42:
after quick discussion with jeremey on the core move pr just going to remove the tests as this may not be adapter specific in anyway

VersusFacit (approved):
I love trimming unneeded code. Thanks for looking into this.

nssalian (approved):
Thanks for checking with core to make sure this is good.

PR #454, opened 2023-02-08
[CT-1970]: convert custom_schema tests to functional

nssalian:
resolves #

### Description

convert custom_schema tests to functional tests

### Checklist


tests/functional/adapter/custom_schema_tests/test_custom_database.py
```diff
@@ -0,0 +1,112 @@
...
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "view_1.sql": _VIEW_1_SQL,
+            "view_2.sql": _VIEW_2_SQL,
+            "view_3.sql": _VIEW_3_SQL,
+        }
+
+    def test_snowflake_override_generate_db_name(self, project):
+        for query in seed_queries:
```
VersusFacit, line 86: nit: why not put this (and whatever else is preliminary) into a fixture called setUp?

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
```diff
@@ -0,0 +1,82 @@
...
+    count(*) as count
+
+from combined
+group by 1
+""".lstrip()
+
+
+class TestCustomProjectSchemaWithPrefix:
+    @pytest.fixture(scope="class")
+    def schema(self):
```
VersusFacit, line 53: What does this do? I suspect this is a noop

VersusFacit (commented):
Some comments, but really clean transfer.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
```diff
@@ -0,0 +1,82 @@
...
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "models": {
+                "schema": "dbt_test"
+            }
+        }
+
+    def test__snowflake__custom_schema_with_prefix(self, project):
+        for query in seed_queries:
```
VersusFacit, line 73: nit: another opportunity for a setup fixture, if nothing else to clean the test function into "what is the meat of the test"

tests/functional/adapter/custom_schema_tests/seeds.py
```diff
@@ -0,0 +1,50 @@
+#
+# Seed queries
```
mikealfare, line 2: Shouldn't these be done with actual seed tables? dbt["seed", "--full-refresh"] will refresh the table, and can be run before each test case.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
nssalian: Yep, removed

tests/functional/adapter/custom_schema_tests/test_custom_database.py
```diff
@@ -0,0 +1,111 @@
...
+from dbt.tests.util import (
+    check_relations_equal,
+    check_table_does_exist,
+    run_dbt
+)
+from tests.functional.adapter.custom_schema_tests.seeds import seed_queries
+
+
+_VIEW_1_SQL = """
+select * from {{ target.schema }}.seed
```
mikealfare, line 12: Since this is creating the thing that you're testing for, I would use the built in ref() function:
```
select * from ref('seed')
```

tests/functional/adapter/custom_schema_tests/seeds.py
nssalian: Not sure I follow. The setUp method should create the tables in the project scope.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
```diff
@@ -0,0 +1,111 @@
...
+    {% else %}
+        {{ target.database }}
+    {% endif %}
+{% endmacro %}
+""".lstrip()
+
+ALT_DATABASE = os.getenv("SNOWFLAKE_TEST_ALT_DATABASE")
+
+class TestOverrideDatabase:
+    @pytest.fixture(scope="class", autouse=True)
```
mikealfare, line 67: There's only one test method in this class, so it doesn't currently matter. But the starting data appears to be written in a way that suggests it should be refreshed for every test, not every class. So the scope should probably be "function" instead of "class", in the event we ever add a second test.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
```diff
@@ -0,0 +1,111 @@
...
+
+select
+    last_name,
+    count(*) as count
+
+from combined
+group by 1
+""".lstrip()
+
+_CUSTOM_DB_SQL = """
```
mikealfare, line 52: This doesn't appear to be used anywhere, unless there are some built-in keywords in dbt that I don't recognize.

nssalian:
```
    def macros(self):
        return {
            "custom_db.sql": _CUSTOM_DB_SQL,
        }
```

tests/functional/adapter/custom_schema_tests/test_custom_database.py
nssalian: This fails the test. Just tried it.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
nssalian: Followed this convention: [setUp](https://github.com/dbt-labs/dbt-core/blob/ccb4fa26cd682812f15d2fd50b37a8c422e092f8/tests/adapter/dbt/tests/adapter/simple_seed/test_seed.py#L45)
If I change it to `function` it doesn't pass when run in parallel.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
VersusFacit: Yeah, and it's  used internally to [do some automagic under the hood](https://docs.getdbt.com/docs/build/custom-databases#generate_database_name)

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
```diff
@@ -0,0 +1,81 @@
+import pytest
+from dbt.tests.util import (
+    check_relations_equal,
+    run_dbt
+)
+from tests.functional.adapter.custom_schema_tests.seeds import seed_queries
+
+_VIEW_1_SQL = """
+select * from {{ target.schema }}.seed
```
mikealfare, line 9: ref('seed')

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
```diff
@@ -0,0 +1,81 @@
...
+    last_name,
+    count(*) as count
+
+from combined
+group by 1
+""".lstrip()
+
+
+class TestCustomProjectSchemaWithPrefix:
+    @pytest.fixture(scope="class", autouse=True)
```
mikealfare: "function"

tests/functional/adapter/custom_schema_tests/test_custom_database.py
nssalian: Added in the next commit.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
nssalian: It's failing the test if I do so.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
mikealfare: Apologies
```
select * from {{ ref('seed') }}
```

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
nssalian: seeing an older version perhaps?

tests/functional/adapter/custom_schema_tests/test_custom_database.py
nssalian: Yep, that's what I did. It fails.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
nssalian:
```
E           dbt.exceptions.TargetNotFoundError: Compilation Error
E             Model 'model.test.view_1' (models/view_1.sql) depends on a node named 'seed' which was not found

```

tests/functional/adapter/custom_schema_tests/seeds.py
mikealfare: You can create a csv from text, like here:
https://github.com/dbt-labs/dbt-redshift/blob/4cc967c966a89a1652ed00d051c7b06d348a0925/tests/functional/adapter/test_snapshot.py#L17

register it as a seed, like here:
https://github.com/dbt-labs/dbt-redshift/blob/4cc967c966a89a1652ed00d051c7b06d348a0925/tests/functional/adapter/test_snapshot.py#L90

and then load it in on setup, like here:
https://github.com/dbt-labs/dbt-redshift/blob/4cc967c966a89a1652ed00d051c7b06d348a0925/tests/functional/adapter/test_snapshot.py#L99

I wrote those tests so that I would never need to reload the seed, so it's just `run_dbt(["seed"])` as a class fixture. But if you wanted to reload after each test method, you could do `run_dbt(["seed", "--full-refresh"])` as a method fixture ("function" in this context) and it will repopulate the seed tables.

nssalian:
I'm not sure whether git is weird. Seeing inconsistencies in the commit files.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
mikealfare: @nssalian - Yeah, I saw that, but I thought it was just creating a macro that never got called.

@VersusFacit - Oh wow, look at that. So `generate_database_name()` is a "keyword". Now I know.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
mikealfare: Interesting. That feels odd. I wonder why we'd need to specify the schema in the model instead of let dbt resolve it.

tests/functional/adapter/custom_schema_tests/test_custom_database.py
nssalian: @mikealfare worth checking, I removed it to make sure it was being used + @VersusFacit 's reference helps to understand this.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
mikealfare: Oh, I bet because the seeds get populated via sql statements instead of as actual seed files. So the models get loaded before the test setup begins. So they don't exist yet, even though they will.

tests/functional/adapter/custom_schema_tests/seeds.py
VersusFacit: That's a good point. I figured there might be some other reason not to use the customary seeds fixture (perhaps order or something like that?). However, Mike, you're right to call out what you have if there's no reason to manually create them via queries.

tests/functional/adapter/custom_schema_tests/test_custom_schema.py
nssalian: I'm curious why the integ tests had it in this way though.

VersusFacit: Honestly, they were like this because integration tests often had no single standard.

colin-rogers-dbt (approved):

PR #503, opened 2023-03-07
added macro for wh selection

### LINKED ISSUE #438: [CT-2011] [PR Review] added macro for wh selection #115
Tracking for:
PR: #115 
Original issue: #103

McKnight-42:
resolves #438 

### Description

take over pr: #115 to get pre-commit issues taken care of and ready for merge

### Checklist


dbt/adapters/snowflake/impl.py
```diff
@@ -73,6 +76,8 @@ def _get_warehouse(self) -> str:
 
     def _use_warehouse(self, warehouse: str):
         """Use the given warehouse. Quotes are never applied."""
+        kwargs = {"warehouse": warehouse}
+        warehouse = self.execute_macro(SNOWFLAKE_WAREHOUSE_MACRO_NAME, kwargs=kwargs)  # type: ignore
```
McKnight-42, line 80: curious if anyone has an idea as to why this causes a 
`dbt/adapters/snowflake/impl.py:80: error: Incompatible types in assignment (expression has type "AttrDict", variable has type "str")  [assignment]` error? seems same as all implementations in `dbt-core` I could find. just curious if there is better path than the ignore

mikealfare: It's silly, but you can remove the type annotation from `warehouse` and `mypy` will be happy; you can also update it to `Any`. The issue is that `self.execute_macro()` lists the type for `kwargs` as `Dict[str, Any]`. `mypy` doesn't like it when you make the type more strict (`str` is a proper subset of `Any`).

.changes/unreleased/Added-20220321-162832.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Added
```
mikealfare, line 1: Is this a supported format? I feel like our process to roll up change logs might not like this. Based on the timestamp, it might be an older format.

McKnight-42: This part might be from before some more recent changes to our changie logic so yeah can remove

dbt/adapters/snowflake/impl.py
McKnight-42: oh that's interesting iI didn't think the type value for the "warehouse" input would bother the previously defined one for `execute_macro` thank you for clearing that up @mikealfare

mikealfare (approved):

manugarri:
@McKnight-42 is there any documentation as to how to use this macro? I assume since its a default macro, it is  enough to use something like this 
```
models:
  +snowflake_warehouse: "ARG_SMALL"    # use the cal the snowflake_warehouse macro with argument `ARG_SMALL` and use the output warehouse for all models in the project...
  my_project:
```

I am testing this by using the macro in a shared dbt package and doesnt seem to work.

Am I understanding this wrong? I cant find any example of how to use the macro and it would be very helpful to us.

McKnight-42:
@manugarri we did some releases yesterday actually https://pypi.org/project/dbt-snowflake/1.5.0b3/ I believe this should be apart of that release

manugarri:
Awesome, great! One question I have, is it not possible to access environment variables inside this macro?

im trying to return different warehouses depending on the snowflake user running dbt, this is what i have so far

```{% macro snowflake_warehouse(warehouse) -%}
    {% set warehouses_specs = {
            "user1": {
                "SMALL": "wh1_small",
                "LARGE": "wh1_large"
            },
            "user2": {
                "SMALL": "wh2_small",
                "LARGE": "wh2_large"
            }
        }
    %}

    {{ return(warehouses_specs[env_var('SNOWFLAKE_USER')][warehouse] ) }}
{%- endmacro %}
```

however when i do dbt run i get these weird error:
`'MacroManifest' object has no attribute 'env_vars'`

How could I read the current_user inside this macro?

McKnight-42:
after some more digging we have decided for now to actually revert https://github.com/dbt-labs/dbt-snowflake/pull/503 out of the prerelease for now as the execute_macro part of code is trying to reference stuff from the manifest before manifest is built and is unintended. sorry for any inconvenience this may cause.

PR #521, opened 2023-03-14
Update create_table_as to use get_select_subquery for column ordering

### LINKED ISSUE #498: [ADAP-349] [Snowflake] Generate order-agnostic constraints ddl when contract: true
Core issue: https://github.com/dbt-labs/dbt-core/issues/7064

Wrap create_table_as user sql in schema-ordered select statement for models with contract: true

```
{% macro <adapter>__create_table_as(temporary, relation, sql) -%}
    ... 
    {% if config.get('contract', False) %}
      {{ get_assert_columns_equivalent(sql) }}
      {{ get_columns_spec_ddl() }}
      {%- set sql = get_select_subquery(sql) %}
    {% endif %}
  as (
    {{ sql }}
  );
{%- endmacro %}
```

We won't need to do this for view materializations as they don't support constraints.

gshank:
resolves #498

### Description

Modifies "create_table_as" to use get_select_subquery to preserve the ordering of columns specified for models with contracts enabled.

### Checklist


tests/functional/adapter/test_constraints.py
```diff
@@ -13,11 +13,16 @@
     id integer not null primary key ,
     color text ,
     date_day date
-) as (
+) as ( select
```
mikealfare, line 16: Maybe I'm reading this wrong, but I don't think this change does anything.

gshank: In that case the order of the columns was the same in SQL and yaml. I changed the test case to make the difference more obvious.

mikealfare: Ah, column order, got it.

mikealfare (approved):

mikealfare (approved):

PR #549, opened 2023-04-04
Adap 18/convert simple copy test

VersusFacit:
### Description

This one was hard because of:
* all the special exceptions to snowflake
* function many of these tests used in the old framework is no longer callable so I had to get creative with manipulating the project YML instead
* a decent amount of effort into trying to manipulate the unique schema fixture with no real success because we don't have an easy entry points into how the profile is constructed (to my knowledge).

### Checklist


tests/integration/simple_copy_test/test_simple_copy.py
```diff
@@ -1,311 +0,0 @@
...
-        self.assertEqual(len(results),  1)
-        results = self.run_dbt()
-        self.assertEqual(len(results),  7)
-
-        self.assertManyTablesEqual(["seed", "view_model", "incremental", "materialized", "get_and_ref"])
-
-        # can't run the test as this one's identifiers will be the wrong case
-
-
-class BaseLowercasedSchemaTest(BaseTestSimpleCopy):
```
VersusFacit, line 163: So, I axed these only after trying to make them work for a while. The problem is that the way we override the scheme here is no longer an option because of how fixture objects are generated at runtime. 

I tried a lot of different hacks for manipulating the state of the project fixture but the connection object is already open. So it's very difficult to manipulate the schema name. I'm not sure how to override it at this point. I did try to override the unit texture but that doesn't really do it either because that's involved before test class.

And last, I wasn't even sure if these are very much worth the trouble anymore since quoting  rules are tested elsewhere

McKnight-42 (approved):

tests/functional/adapter/simple_copy/test_simple_copy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def test_simple_copy(self, project):
+        super().test_simple_copy(project)
+        run_dbt(['test'])
+
+
+class TestSimpleCopyBaseQuotingOff(SimpleCopyBase):
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {"quoting": {"identifier": False}}
```
nssalian, line 39: nit: maybe formatting would make this cleaner or did it pass through?

nssalian (approved):

VersusFacit:
Seeing as I'm going to reformat this whole repo as soon as Neelesh finishes up the snapshot test conversion, I'm going to merge things as is

PR #572, opened 2023-04-17
Winkler/show more objects

### LINKED ISSUE #571: [ADAP-461] [Enhancement] Enable listing >10K objects in a schema
### Current Behavior

When a [target schema contains more than 10,000 tables + views](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/include/snowflake/macros/adapters.sql#L153), dbt throws a compilation exception. 

### Expected Behavior

Make dbt additionally robust to this situation, such that a compilation error is not triggered at an arbitrary (though imposed by Snowflake) threshold. See [here](https://docs.snowflake.com/en/sql-reference/sql/show-objects) for more details on the Snowflake limits.

### Steps To Reproduce

`dbt run` in an environment with a target schema that contains more than 10K tables / views. 

### Relevant log output

_No response_

### Environment

_No response_

### Additional Context

Based on the [diff](https://github.com/dbt-labs/dbt-core/pull/2322/files#diff-4e52d38c66fb471207fa26868ecc11715965558734704859a52b6bcfbc5c86cfL96-L111) from when we changed the approach here last, could we:

1. Run `SHOW TERSE OBJECTS`
2. If we see 10K results
3. Fall back to querying the information schema

?

matt-winkler:
resolves #571 

### Description

1. Run SHOW TERSE OBJECTS
2. If we see 10K results
3. Fall back to querying the information schema

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-snowflake/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [X] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-snowflake/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

### DIFF
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -173,7 +173,7 @@
   {%- endset -%}
 
   {%- set result = run_query(sql) -%}
-  {% set maximum = 1 %}
+  {% set maximum = 10000 %}
   {% if (result | length) >= maximum %}
     {% set msg %}
       Too many objects in schema  {{ schema_relation }}! dbt can only get
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -155,14 +155,41 @@
     show terse objects in {{ schema_relation }}
   {%- endset -%}
 
+  {%- set fallback_sql -%}
+     select
+      created as "created_on",
+      table_name as "name",
+      case when table_type = 'BASE TABLE' then 'table'
+           when table_type = 'VIEW' then 'view'
+           when table_type = 'MATERIALIZED VIEW' then 'materializedview'
+           when table_type = 'EXTERNAL TABLE' then 'external'
+           else table_type
+      end as "kind",
+      table_catalog as "database_name",
+      table_schema as "schema_name"
+    from  information_schema.tables
+    where table_schema ilike '{{ schema_relation.schema.lower() }}'
+      and table_catalog ilike '{{ schema_relation.database.lower() }}'
+  {%- endset -%}
+
   {%- set result = run_query(sql) -%}
-  {% set maximum = 10000 %}
+  {% set maximum = 1 %}
   {% if (result | length) >= maximum %}
     {% set msg %}
       Too many objects in schema  {{ schema_relation }}! dbt can only get
-      information about schemas with fewer than {{ maximum }} objects.
+      information about schemas with fewer than {{ maximum }} objects using
+      the "show terse objects" command.
+
+      Falling back to querying the Snowflake information schema directly.
+
+      To improve performance, we recommend confirming all tables and
+      views in {{ schema_relation}} are current.
+
     {% endset %}
-    {% do exceptions.raise_compiler_error(msg) %}
+
+    {{ log(msg, info=true) }}
+    {%- set result = run_query(fallback_sql) %}
+
   {% endif %}
   {%- do return(result) -%}
 {% endmacro %}
--- a/.changes/unreleased/Features-20230417-173251.yaml
+++ b/.changes/unreleased/Features-20230417-173251.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: enable showing more than 10K objects in a schema
+time: 2023-04-17T17:32:51.159541-06:00
+custom:
+  Author: matt-winkler
+  Issue: "571"


dbt/include/snowflake/macros/adapters.sql
```diff
@@ -155,14 +155,41 @@
...
+      case when table_type = 'BASE TABLE' then 'table'
+           when table_type = 'VIEW' then 'view'
+           when table_type = 'MATERIALIZED VIEW' then 'materializedview'
+           when table_type = 'EXTERNAL TABLE' then 'external'
+           else table_type
+      end as "kind",
+      table_catalog as "database_name",
+      table_schema as "schema_name"
+    from  information_schema.tables
+    where table_schema ilike '{{ schema_relation.schema.lower() }}'
```
mikealfare, line 171: Would changing these filters from `ilike` to equality help with the performance issues?

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -155,14 +155,41 @@
     show terse objects in {{ schema_relation }}
   {%- endset -%}
 
+  {%- set fallback_sql -%}
+     select
+      created as "created_on",
+      table_name as "name",
+      case when table_type = 'BASE TABLE' then 'table'
+           when table_type = 'VIEW' then 'view'
+           when table_type = 'MATERIALIZED VIEW' then 'materializedview'
```
mikealfare, line 164: Should this be `materializedview` or `materialized_view`? I think we use both for different reasons (I don't know why and wish we didn't), so we should make sure it's the right one.

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -155,14 +155,41 @@
     show terse objects in {{ schema_relation }}
   {%- endset -%}
 
+  {%- set fallback_sql -%}
+     select
+      created as "created_on",
+      table_name as "name",
+      case when table_type = 'BASE TABLE' then 'table'
```
mikealfare, line 162: Minor nit, and not sure if this works in snowflake sql, but these type of case statements can also usually be written this way:
```sql
case table_type
    when 'BASE TABLE' then 'table'
    when 'VIEW' then 'view'
    when 'MATERIALIZED VIEW' then 'materializedview'
    when 'EXTERNAL TABLE' then 'external'
    else table_type
end as "kind"
```

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -155,14 +155,41 @@
...
     {% set msg %}
       Too many objects in schema  {{ schema_relation }}! dbt can only get
-      information about schemas with fewer than {{ maximum }} objects.
+      information about schemas with fewer than {{ maximum }} objects using
+      the "show terse objects" command.
+
+      Falling back to querying the Snowflake information schema directly.
+
+      To improve performance, we recommend confirming all tables and
+      views in {{ schema_relation}} are current.
```
mikealfare, line 186: I think this needs to be `{{ schema_relation }}`? Otherwise the missing space throws an error.

mikealfare:
This looks like an interesting one to test, lol

dbt/include/snowflake/macros/adapters.sql
matt-winkler: Hey @mikealfare thanks for reviewing this. I learned late yesterday that in fact we can paginate the `show terse objects` query on snowflake, so shifted direction there in working with @dbeatty10.

dbt/include/snowflake/macros/adapters.sql
matt-winkler: Hey @mikealfare thanks for reviewing this. I learned late yesterday that in fact we can paginate the `show terse objects` query on snowflake, so shifted direction there in working with @dbeatty10.

dbt/include/snowflake/macros/adapters.sql
matt-winkler: Hey @mikealfare thanks for reviewing this. I learned late yesterday that in fact we can paginate the `show terse objects` query on snowflake, so shifted direction there in working with @dbeatty10.

dbt/include/snowflake/macros/adapters.sql
matt-winkler: Hey @mikealfare thanks for reviewing this. I learned late yesterday that in fact we can paginate the `show terse objects` query on snowflake, so shifted direction there in working with @dbeatty10.

matt-winkler:
I could use some help in a couple areas on this @mikealfare @dbeatty10:

- Ensuring what I have here is performant. In particular do you see any major red flags with holding (up to) 10 10k row agate tables in memory and then merging them?
- Code readability. I worked around a few jinja limitations here, namely the inability to update a variable's value in the context of a for loop when that variable is defined outside the loop. Someone smarter than me may have a better approach.
- How to test this.

mikealfare:
You could also create the loop in the `models()` fixture:
```python
def models():
    my_models = {"my_model_0.sql": "{{ config(materialized='table') }} select 1 as id;"}
    for view in range(1,15000):
        my_models.update({f"my_model_{view}.sql": "select * from {{ ref('my_model_{view-1}') }};"})
    return my_models

**Comment by matt-winkler** [comment]
Hey @mikealfare thanks for that example. I borrowed from it and got a successful test run on my local 👍 . 



There are a few modifications in this version vs. yours: 

```
class TestListRelationsWithoutCaching:
    def models(self):
        my_models = {"my_model_0.sql": _TABLE_1_SQL}
        for view in range(1, 10100):
            my_models.update({f"my_model_{view}.sql": _VIEW_X_SQL})

        return my_models
```

HOWEVER, it took about 2 hours for this to run on Snowflake. Is that something we are concerned about? Should we perhaps override the `list_relations_without_caching` macro to use a lower count of objects before triggering the loop just for the purpose of integration testing? Use a dedicated warehouse with higher thread count?

File: tests/functional/adapter/test_list_relations_without_caching.py [hunk 5]
```diff
@@ -0,0 +1,105 @@
...
+#   into a fresh testing schema
+# - accordingly, we create a smaller set of views and test the looping iteration logic in
+#   smaller chunks
+
+NUM_VIEWS = 100
+NUM_EXPECTED_RELATIONS = 1 + NUM_VIEWS
+
+# these are arguments to the macro which lists all relations in a target schema
+MAX_ITER = 11
+MAX_RESULTS_PER_ITER = 10
```
VersusFacit, line 20: unused? Was this intended to be used at line 34?

VersusFacit:
> Hey @mikealfare thanks for that example. I borrowed from it and got a successful test run on my local 👍 .
> 
> There are a few modifications in this version vs. yours:
> 
> ```
> class TestListRelationsWithoutCaching:
>     @pytest.fixture(scope="class")
>     def models(self):
>         my_models = {"my_model_0.sql": _TABLE_1_SQL}
>         for view in range(1, 10100):
>             my_models.update({f"my_model_{view}.sql": _VIEW_X_SQL})
> 
>         return my_models
> ```
> 
> HOWEVER, it took about 2 hours for this to run on Snowflake. Is that something we are concerned about? Should we perhaps override the `list_relations_without_caching` macro to use a lower count of objects before triggering the loop just for the purpose of integration testing? Use a dedicated warehouse with higher thread count?

This is now out of date right? No longer doing 10k, and in fact only about 100.

tests/functional/adapter/test_list_relations_without_caching.py
```diff
@@ -0,0 +1,105 @@
...
+    def macros(self):
+        return {
+            "validate_list_relations_without_caching.sql": MACROS__VALIDATE__SNOWFLAKE__LIST_RELATIONS_WITHOUT_CACHING
+        }
+
+    def test__snowflake__list_relations_without_caching(self, project):
+        # purpose of the first run is to create the replicated views in the target schema
+        _ = run_dbt(["run"])
+
+        # there is probably a better way to get the database and schema than this
```
VersusFacit, line 84: nah, this is idiomatic enough. Better this than `project.adapter` params

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,21 +150,83 @@
...
-    {% set msg %}
-      Too many objects in schema  {{ schema_relation }}! dbt can only get
-      information about schemas with fewer than {{ maximum }} objects.
-    {% endset %}
-    {% do exceptions.raise_compiler_error(msg) %}
+  {%- set n = (result | length) -%}
+  {%- set counts_array = [n] -%}
+  {%- set paginated_results_array = [] -%}
+
+  {% if n >= max_results_per_iter %}
```
VersusFacit, line 166: this is such a long `if`, at line 224 (I think?) it be great to pull this into its own function with a descriptive name like `get_paginated_relations_array` (_vel sim._) and just have a function call. 

If for whatever reason that's not possible, you could also leave a comment online 224. My first suggestion is much preferred though

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,21 +150,83 @@
...
   {%- set result = run_query(sql) -%}
-  {% set maximum = 10000 %}
-  {% if (result | length) >= maximum %}
-    {% set msg %}
-      Too many objects in schema  {{ schema_relation }}! dbt can only get
-      information about schemas with fewer than {{ maximum }} objects.
-    {% endset %}
-    {% do exceptions.raise_compiler_error(msg) %}
+  {%- set n = (result | length) -%}
+  {%- set counts_array = [n] -%}
```
VersusFacit, line 163: I'm feeling kind of dumb right now. What is the reason for this being a list? Why can't this just be a variable that we're adding two progressively as opposed to a list object that we're adding new results too. I feel like making this just a simple accumulator would simplify some of the control flow in this function

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,21 +150,83 @@
...
+
+    {% for _ in range(1, (max_iter + 1) ) %}
+
+      {#
+        terminating condition: At some point the user needs to be reasonable with how
+        many objects are contained in their schemas
+      #}
+
+      {%- if loop.index == max_iter -%}
+
```
VersusFacit, line 182: nit: nix some of the whitespace for this if block through 189

dbt/include/snowflake/macros/adapters.sql
VersusFacit: Oh, that's right Ginger is a pain about this. I've got a work around for you:

```
{% set ns = namespace(sum=0) %}
{% for value in values %}
  {%- set ns.sum = ns.sum + value %}
{% endfor %}
The sum of the values is: {{ ns.sum }}
```

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,21 +150,83 @@
...
-    {% endset %}
-    {% do exceptions.raise_compiler_error(msg) %}
+  {%- set n = (result | length) -%}
+  {%- set counts_array = [n] -%}
+  {%- set paginated_results_array = [] -%}
+
+  {% if n >= max_results_per_iter %}
+
+    {# cant update a variable in the context of a loop, so sticking in an array instead #}
+    {% set last_values_array = [] %}
```
VersusFacit, line 169: If possible, use a namespace object call here too (I think you can use the other one).

VersusFacit (commented):
I'm so happy that we found a way to continue to use objects instead of having to deal with the information schema. 

I'd really like to see this re-implemented with a namespace object almost-normal-imperative style to stop tracking values using lists (it's just so complex relative to what it is in practice). Pagination logic itself seems sound though. 

I also think that the task you've written is totally appropriate given snowflakes on restrictions and the absurdity of creating 30,000 objects for a small integration test

tests/functional/adapter/test_list_relations_without_caching.py
matt-winkler: Awesome. I think that comment was outdated from me parsing them out of a run result but then found this somewhere else.

dbt/include/snowflake/macros/adapters.sql
matt-winkler: super helpful, thanks

dbt/include/snowflake/macros/adapters.sql
matt-winkler: done

dbt/include/snowflake/macros/adapters.sql
matt-winkler: done

tests/functional/adapter/test_list_relations_without_caching.py
matt-winkler: I removed these. Let me know if they should be added back / parameterized further.

matt-winkler:
@VersusFacit thanks so much for the great review and pro tip on using namespace variables which made this so much easier. I believe this is ready for another round of review with additional macro abstraction, tests, and a tweak to the termination logic as a result.

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,21 +150,93 @@
...
-  {% set maximum = 10000 %}
-  {% if (result | length) >= maximum %}
-    {% set msg %}
-      Too many objects in schema  {{ schema_relation }}! dbt can only get
-      information about schemas with fewer than {{ maximum }} objects.
-    {% endset %}
-    {% do exceptions.raise_compiler_error(msg) %}
+
+  {%- set n = (result | length) -%}
+  {%- set total_count = namespace(sum=n) -%}
```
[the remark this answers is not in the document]
VersusFacit: On the contrary, I think you did exactly the right thing with that naming convention.

I'm more confused about how you've used these names in the code. If I’m reading you right, line 222 should be `{% if total_count >= max_results_per_iter %}` (instead of `{% if n >= max_results_per_iter %}`) and all other references to total_count should be removed from `snowflake__get_paginated_relations_array` both its header (153), call (229), and implementation (192).

PR #659, opened 2023-06-14
ADAP-544: Add configuration change options

### LINKED ISSUE #603: [ADAP-544] [Feature] Support Dynamic Tables as a materialization (following base Materialized View ergonomics)
For more context about Dynamic Tables (DTs), see [this post](https://www.snowflake.com/blog/dynamic-tables-delivering-declarative-streaming-data-pipelines/).

## `materialized: dynamic_table`

This issue is the local implementation of the core feature: [Let's add Materialized View as a materialization, finally](https://github.com/dbt-labs/dbt-core/issues/6911). 
The original spec is the baseline. In every aspect it's the same feature, with the exceptions described below.

### Alias

The local alias for this materialization is `dynamic_table`, not `materialized_view`. The later is not supported and should return an error (to avoid confusion with Snowflake's MVs)

### Schema management

Metadata caching/querying should be updated to add the following sources in order to support Dynamic Tables:
- table level via `SHOW DYNAMIC TABLES` - [doc](https://docs.snowflake.com/en/LIMITEDACCESS/show-dynamic-tables)
- column level via `DESCRIBE DYNAMIC TABLE` - [doc](https://docs.snowflake.com/LIMITEDACCESS/desc-dynamic-table)

### Configurations

Available parameters:
  - ~~LAG~~ TARGET_LAG : string (`<num> { seconds | minutes | hours | days }`)
  - WAREHOUSE : string
  - REFRESH_STRATEGY : {`no_wait`,`refreshed`}
  - From models: query_tag

Refresh strategy is described below.

**Out of scope** for the moment (not clear yet if applicable): [copy_grants](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants), [persist_docs](https://docs.getdbt.com/reference/resource-configs/persist_docs), [clustering](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-table-clustering) management

### Lifecycles

- Lifecycle (`dbt run`) : 
  - **If** necessary backup (exists with different materialization)
  - **If** necessary `CREATE OR REPLACE DYNAMIC TABLE` (doesn't exist) and apply *refresh_strategy* **Else** 
    - Check status (see below)
      - **If** `FAILED` then `CREATE OR REPLACE DYNAMIC TABLE` and apply *refresh_strategy* **Else** 
        - **If** the options of the object in the database are different from the current model, depending on `on_configuration_change`
        - `skip` : Do nothing, aka stage changes until the next full refresh
        - `fail` : Fail the run
        - `apply` : [default] Apply the changes
 - Lifecycle (`dbt run --full-refresh`) :
   - **If** necessary backup (exists with different materialization ~~OR exists and full-refresh~~)
   - `CREATE OR REPLACE DYNAMIC TABLE` and apply *refresh_strategy*

### Status check

The refresh status of a DT is obtained via the following query:

```SQL 
SELECT TOP 1 state
FROM TABLE(INFORMATION_SCHEMA.DYNAMIC_TABLE_REFRESH_HISTORY(NAME => '...'))
ORDER BY refresh_version DESC;
```

The return code is a string [among](https://docs.snowflake.com/LIMITEDACCESS/dynamic_table_refresh_history#output):

> SCHEDULED, EXECUTING, SUCCEEDED, FAILED, CANCELLED, or SKIPPED

### Refresh strategy

To ensure downstream models are able to query the DT in a proper state, dbt may have to wait for the DT to be successfully refreshed before moving on. The `refresh_strategy` setting offers 2 options:

|Refresh strategy|Behavior|
|-|-|
|`no_wait`|default: dbt will not wait, it will proceed as soon as the DDL statement is issued|
|`refresh`|optional: dbt will wait until the DT is in the `SUCCEEDED` state, polling every few seconds its status|

In order to avoid waiting indefinitely, in `refresh` mode, the polling will stop and the **run will fail** when:
- after the TARGET_LAG period has lapsed, any state other than `SCHEDULED, EXECUTING` is returned
- after 2 TARGET_LAG periods have lapsed

mikealfare:
resolves #603

### Description

Implement dynamic tables as a materialization option for Snowflake.

Workflow:
- Refresh existing dynamic table with no changes
- Accept and process changes to a dynamic table:
  - `target_lag`: alter dynamic table in place
  - `warehouse`: full refresh

Configuration:
- `on_configuration_change`: determines the action to take when changes are detected:
  - `apply`: full refresh on warehouse changes, alter on target_lag changes
  - `continue`: soft fail, skip the processing of this model, but process downstream
  - `fail`: fail entire run

### Checklist


dbt/adapters/snowflake/relation.py
```diff
@@ -14,17 +31,18 @@ class SnowflakeRelationType(StrEnum):
...
 @dataclass(frozen=True, eq=False, repr=False)
 class SnowflakeRelation(BaseRelation):
+    # we need to overwrite the type annotation for `type` so that mashumaro reads the class correctly
     type: Optional[SnowflakeRelationType] = None  # type: ignore
+
+    include_policy: SnowflakeIncludePolicy = field(
+        default_factory=lambda: SnowflakeIncludePolicy()
+    )
     quote_policy: SnowflakeQuotePolicy = field(default_factory=lambda: SnowflakeQuotePolicy())
+    relation_configs = {
```
VersusFacit, line 43: Why is this plural if there is only one? We got more coming?

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
     @classproperty
     def DynamicTable(cls) -> str:
         return str(SnowflakeRelationType.DynamicTable)
+
+    @classproperty
+    def get_relation_type(cls) -> Type[SnowflakeRelationType]:
+        return SnowflakeRelationType
+
+    @classmethod
+    def from_runtime_config(cls, runtime_config: RuntimeConfigObject) -> RelationConfigBase:
```
VersusFacit, line 60: What does this function mean? I tried figuring that out from reading where it is invoked. That said, it's not immediately clear. You want the model from the runtime config which the materialization invokes? Just a lot of indirection here.

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
...
+        model_node: ModelNode = runtime_config.model
+        relation_type = SnowflakeRelationType(model_node.config.materialized)
+
+        if relation_config := cls.relation_configs.get(relation_type):
```
VersusFacit, line 64: Oh this operator. Oh good.

What's your intended semantics? Non-`Optional` assignment semantics frighten me.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+                validation_error=DbtRuntimeError(
+                    f"Warehouse should only be altered but {self.action} was received."
+                ),
+            ),
+        }
+
+
+@dataclass
+class SnowflakeDynamicTableConfigChangeset:
+    target_lag: Optional[SnowflakeDynamicTableTargetLagConfigChange] = None
```
VersusFacit, line 167: Can't remember. Could you just have the vars without the `= None` and get the same thing? Are you just erring to explicit? Is one considered cleaner? Can I find a way to ask more questions about this one line?

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
...
+        cls,
+        dynamic_table: SnowflakeDynamicTableConfig,
+        relation_results: RelationResults,
+    ) -> Optional[SnowflakeDynamicTableConfigChangeset]:
+        config_changeset = SnowflakeDynamicTableConfigChangeset()
+
+        existing_dynamic_table = SnowflakeDynamicTableConfig.from_relation_results(
+            relation_results
+        )
+        assert isinstance(existing_dynamic_table, SnowflakeDynamicTableConfig)
```
VersusFacit, line 82: Leftover `assert`? Should it be an exception?

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
...
+            return relation_config.from_model_node(model_node)
+
+        raise DbtRuntimeError(
+            f"from_runtime_config() is not supported for the provided relation type: {relation_type}"
+        )
+
+    @classmethod
+    def dynamic_table_config_changeset(
+        cls,
+        dynamic_table: SnowflakeDynamicTableConfig,
```
VersusFacit, line 74: So this is pointing to what we're creating? Should this be like `target_` prefixed or something like that to highlight this is the to-be-built copy of an existing table?

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
...
+
+        if dynamic_table.warehouse != existing_dynamic_table.warehouse:
+            config_changeset.warehouse = SnowflakeDynamicTableWarehouseConfigChange(
+                action=RelationConfigChangeAction.alter,
+                context=dynamic_table.warehouse,
+            )
+
+        if config_changeset.has_changes:
+            return config_changeset
+        return None
```
VersusFacit, line 98: nit: put an `else` ;p

dbt/adapters/snowflake/relation.py
```diff
@@ -33,3 +51,48 @@ def is_dynamic_table(self) -> bool:
...
+
+        if relation_config := cls.relation_configs.get(relation_type):
+            return relation_config.from_model_node(model_node)
+
+        raise DbtRuntimeError(
+            f"from_runtime_config() is not supported for the provided relation type: {relation_type}"
+        )
+
+    @classmethod
+    def dynamic_table_config_changeset(
```
VersusFacit, line 72: So this function is building the changeset between the new and existing version of the table? I find the name doesn't capture these semantics, and it would help self-document

.changes/unreleased/Features-20230620-224944.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Add support for dynamic tables
```
VersusFacit, line 2:
```suggestion
body: Add support for dynamic tables, the Snowflake implementation of materialized views
```
or something like that 🤷‍♀️

dbt/adapters/snowflake/relation_configs/base.py
```diff
@@ -0,0 +1,50 @@
...
+    @classmethod
+    def parse_relation_results(cls, relation_results: RelationResults) -> dict:
+        raise NotImplementedError(
+            "`parse_relation_results()` needs to be implemented on this RelationConfigBase instance"
+        )
+
+    @staticmethod
+    def _render_part(component: ComponentName, value: str) -> Optional[str]:
+        if SnowflakeIncludePolicy().get_part(component):
+            if SnowflakeQuotePolicy().get_part(component):
```
VersusFacit, line 47: Is the QuotePolicy intended to be subordinate to the IncludePolicy?

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
 
     name: str
+    schema_name: str
+    database_name: str
     query: str
     target_lag: SnowflakeDynamicTableTargetLagConfig
     warehouse: str
+
+    @property
+    def path(self) -> str:
```
VersusFacit, line 45: Howabout relation path?

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+        return ".".join(
+            part for part in [self.database_name, self.schema_name, self.name] if part is not None
```
VersusFacit, line 47: So what happens if `self.schema_name` is `None` but the other's aren't?

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+        )
+
+    @property
+    def schema_path(self) -> str:
+        return ".".join(
+            part for part in [self.database_name, self.schema_name] if part is not None
```
VersusFacit, line 53: ditto to above

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
```diff
@@ -93,27 +92,31 @@
 {% endmacro %}
 
 
-{% macro dynamic_table_execute_no_op(target_relation) %}
+{% macro dynamic_table_execute_no_op(dynamic_table) %}
```
VersusFacit, line 95: `target_dynamic_table` be a good name perchance?

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
```diff
@@ -93,27 +92,31 @@
...
     ) %}
 {% endmacro %}
 
 
-{% macro dynamic_table_execute_build_sql(build_sql, existing_relation, target_relation, post_hooks) %}
+{% macro dynamic_table_execute_build_sql(build_sql, dynamic_table, existing_relation, post_hooks) %}
 
     {% set grant_config = config.get('grants') %}
 
+    {% set original_query_tag = set_query_tag() %}
```
VersusFacit, line 109: Where did this come from/why? I don't see anything in your issue spec describing it!

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
```diff
@@ -1,46 +1,96 @@
...
 ) -%}
-    {{- log('Applying ALTER to: ' ~ relation) -}}
-    {{- snowflake__get_replace_dynamic_table_as_sql(relation, sql, existing_relation, backup_relation, intermediate_relation) -}}
+    {{- log('Applying ALTER to: ' ~ dynamic_table.path) -}}
+
+    {% if configuration_changes.requires_full_refresh %}
+
+        {{ snowflake__replace_dynamic_table_sql(dynamic_table, existing_relation, backup_relation, intermediate_relation) }}
+
+    -- otherwise apply individual changes as needed
```
VersusFacit, line 14:
```suggestion
    -- otherwise apply the changes collected in configuration_changes
```
or whatever would be most descriptive.

Also, there is an argument a good in-code comment says why rather than what so maybe just a rewrite to
```
    -- a full refresh does not occur _<blah blah>___ 
```

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
```diff
@@ -1,46 +1,96 @@
...
+    {{- log('Applying CREATE to: ' ~ dynamic_table.path) -}}
 
-    create or replace dynamic table {{ relation }}
-        lag = '{{ config.get("target_lag") }}'
-        warehouse = {{ config.get("warehouse") }}
-        as ({{ sql }})
+    create or replace dynamic table {{ dynamic_table.path }}
+        target_lag = '{{ dynamic_table.target_lag }}'
+        warehouse = {{ dynamic_table.warehouse }}
+        as ({{ dynamic_table.query }})
```
VersusFacit, line 37: Okay so I _think_ this is one of those places that without spacing to

```sql
as (
  {{ dynamic_table.query }}
)
```
would result in bugs. We've seen this in RS (or whichever db 🙈 ). Did you prescrub out comments? If not, we should probably code defensively here.

dbt/adapters/snowflake/relation_configs/target_lag.py
```diff
@@ -26,3 +39,101 @@ class SnowflakeDynamicTableTargetLagConfig(RelationConfigBase):
...
+                validation_check=self.duration is not None and self.period is not None,
+                validation_error=DbtRuntimeError(
+                    "Target lag for a materialized view requires both a duration and a period."
+                ),
+            ),
+            RelationConfigValidationRule(
+                validation_check=(
+                    (
+                        self.duration >= 1
+                        and self.period != SnowflakeDynamicTableTargetLagPeriod.seconds
```
VersusFacit, line 56: `seconds` -> `minute`?

VersusFacit: And could someone say 0 hours by chance? 🙊

dbt/adapters/snowflake/relation_configs/target_lag.py
```diff
@@ -26,3 +39,101 @@ class SnowflakeDynamicTableTargetLagConfig(RelationConfigBase):
...
+            kwargs_dict.update({"duration": int(duration)})
+
+        if period := config_dict.get("period"):
+            kwargs_dict.update({"period": SnowflakeDynamicTableTargetLagPeriod(period)})
+
+        target_lag: "SnowflakeDynamicTableTargetLagConfig" = super().from_dict(kwargs_dict)  # type: ignore
+        return target_lag
+
+    @classmethod
+    def parse_model_node(cls, model_node: ModelNode) -> dict:
```
VersusFacit, line 84: I don't understand this paradigm you've made but I want to! Who is calling this? Why is the name so generic when the result is so specific to the class? (I see you do this elsewhere)

dbt/adapters/snowflake/relation_configs/target_lag.py
```diff
@@ -26,3 +39,101 @@ class SnowflakeDynamicTableTargetLagConfig(RelationConfigBase):
...
+        target_lag: str = model_node.config.extra.get("target_lag")
+        try:
+            duration, period = target_lag.split(" ")
+        except IndexError:
```
VersusFacit, line 88: Should also catch the `AttributeError` when `target_lag` is `None` (since `get` can do that) unless that's guarded against elsewhere.

VersusFacit: It might be in `dbt/adapters/snowflake/relation_configs/dynamic_table.py` around line `78`? Still, if you ever instantiate this without the `AttributeError` check, you'll get an exception.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+            "schema_name": model_node.schema,
+            "database_name": model_node.database,
+            "query": model_node.compiled_code,
+            "target_lag": SnowflakeDynamicTableTargetLagConfig.parse_model_node(model_node),
+            "warehouse": model_node.config.extra.get("warehouse"),
+        }
+        return config_dict
+
+    @classmethod
+    def parse_relation_results(cls, relation_results: RelationResults) -> dict:
```
VersusFacit, line 99: Where does this method get used? For output or something else?

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
```diff
@@ -1,46 +1,96 @@
...
+            in schema {{ dynamic_table.schema_path }}
+        ;
+        select
+            "name",
+            "schema_name",
+            "database_name",
+            "text",
+            "target_lag",
+            "warehouse"
+        from table(result_scan(last_query_id()))
```
VersusFacit, line 90: concurrency nightmare possible? Maybe.....we could do a dynamic query call? Or are you confident this is thread-safe/otherwise-collision-safe?

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
```diff
@@ -1,46 +1,96 @@
-{% macro snowflake__get_alter_dynamic_table_as_sql(
```
VersusFacit, line 1: Is this file a bunch of helper DDL for the snowflake methods for building out this table? If I'm right, would love to see that documented at filetop. If I'm wrong, another reason to document our intentions 😁

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+                )
+            ;
+
+        Returns: the `select ...` statement, for example:
+
+            select * from my_base_table
+
+        """
+        open_paren = query.find("as (") + len("as (")
+        close_paren = query.rindex(")")
```
VersusFacit, line 139: Damn, that's stressful to me, but it does seem right!

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
...
+        return query[open_paren:close_paren].strip()
+
+
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class SnowflakeDynamicTableWarehouseConfigChange(
+    RelationConfigChange, RelationConfigValidationMixin
+):
+    context: str
```
VersusFacit, line 147: Where does this get used?

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
```diff
@@ -46,39 +45,39 @@
...
     -- determine the scenario we're in: create, full_refresh, alter, refresh data
     {% if existing_relation is none %}
-        {% set build_sql = snowflake__get_create_dynamic_table_as_sql(target_relation, sql) %}
+        {% set build_sql = snowflake__create_dynamic_table_sql(dynamic_table) %}
     {% elif full_refresh_mode or not existing_relation.is_dynamic_table %}
-        {% set build_sql = snowflake__get_replace_dynamic_table_as_sql(target_relation, sql, existing_relation, backup_relation, intermediate_relation) %}
+        {% set build_sql = snowflake__replace_dynamic_table_sql(dynamic_table, existing_relation, backup_relation, intermediate_relation) %}
     {% else %}
 
         -- get config options
```
VersusFacit, line 67: Dangling/incomplete comment?

VersusFacit (commented):
WOW! That's a lot of code! Still, all the boilerplate you wrote made it pretty starightforward to read so kudos.

.changes/unreleased/Features-20230620-224944.yaml
mikealfare: Snowflake also has materialized views. There is a desire to not use "materialized view" in any of the user facing components in order to reduce confusion.

dbt/adapters/snowflake/relation.py
mikealfare: Yes, the idea is that if this new approach holds water, then we would eventually migrate all materializations to this approach as needed. Also, some of the methods that use this below would eventually move up into `dbt-core`. This would be the way to "register" your relation config with a relation type, so to speak.

dbt/adapters/snowflake/relation.py
mikealfare: I definitely need a docstring here; I'll add it. The idea is to produce a `RelationConfig` from the `RuntimeConfigObject`; in this case it would produce a `SnowflakeDynamicTableConfig` instance. `RuntimeConfigObject` contains all of the information needed to create a `SnowflakeDynamicTableConfig`, plus a lot more. There are two key differences. The latter is in the parlance of the Snowflake object, and not the generic `RuntimeConfigObject`. And producing the latter provides a much better place for validation checks and structural checks; we don't need to do it at the `SnowflakeRelation` level or in the jinja.

dbt/adapters/snowflake/relation.py
mikealfare: In this case we control the value, and it can't be `None` because relation_type is a `Enum`. But I do explicitly use that in all of the config parsers. Passing `None` is effectively not passing it at all. I think I covered the scenarios there, where `None` is not actually a valid value for that config parameter. But agreed, we need to watch out for that and check for the existence of the key in those scenarios. I could also make the argument that we don't allow that value in our interface.

dbt/adapters/snowflake/relation.py
mikealfare: I went around the block on naming this and I agree that it's not super intuitive. I'll see if I can come up with something.

For context, here was my thought process leading up to this name. At this point in the process, we know that there's an existing database object and an object in the config that points to that database object, which may or may not have changes. Both things are really the same thing, the same dynamic table. We're merely taking another snapshot and then determining if the state of the object (minus data) in the config matches the already existing object in the database. Since they both represent the same dynamic table at two points in time, we're effectively determining the changeset between those two points in time on the same dynamic table; hence `dynamic_table_config_changeset()`. This makes me wonder if maybe the method name isn't that bad, but the arguments should be updated, as you point out below.

dbt/adapters/snowflake/relation.py
mikealfare: I'm definitely focusing on the new snapshot, which lives in the user config, as the "primary" view of the dynamic table. I did that because that's where I am when I'm running this, in the future looking back at the past. However, as noted above, this should probably change. I also don't like `relation_results` since it's very vague. This was the result of several back and forth's in reviews. I think I just might bite the bullet, and do the parsing for the `existing_dynamic_table` outside of this method as well, and then wind up with `existing_dynamic_table` and `new_dynamic_table` as the arguments, which is more clear.

dbt/adapters/snowflake/relation.py
mikealfare: This is to make the linter happy. `from_relation_results()` returns a generic `RelationConfigBase()` because it's on the base class. The base class has no platform-specific attributes like `target_lag`, `warehouse`, etc. Instead of doing a `# type: ignore` and dealing with the warnings in the IDE, I solve both with this assert.

dbt/adapters/snowflake/relation.py
mikealfare: ::gets on soap box and raises pinky:: My understanding is that it's more idiomatic to return the `None` outside of the `else` block since the `else` block isn't doing anything. That's how I've generally seen it, so the `else: return None` construct looks strange to me, as if we could get past it somehow and not return anything (because it's indented). A compromise could be:
```python
return config_changeset if config_changeset.has_changes else None
```

dbt/adapters/snowflake/relation_configs/base.py
mikealfare: If we're not including that part, then we don't care about the quote policy for that part. But if we are including that part, then we need to see if we're quoting it.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -21,6 +35,147 @@ class SnowflakeDynamicTableConfig(RelationConfigBase):
     """
 
     name: str
+    schema_name: str
```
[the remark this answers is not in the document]
mikealfare: True, I did that in cases when there are defaults because passing in `my_arg=None` does not result in the default for `my_arg`, it results in `None`. However, I don't have defaults. I'll remove those guards below.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: Since this will always be used in the form `dynamic_table.path`, I'm not keen on `dynamic_table.relation_path` since it feels like doublespeak. I chose `path` because I believe we call it that in `dbt-core` somewhere. However, `path` is generic; what this really is is the "fully qualified path". And since we're in `relation_config` land, we should be talking in database parlance. I'll update it to `fully_qualified_path`, but let me know your thoughts on that name.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: lol, I thought the same thing, but this is how we roll in `dbt-core`. fwiw, those three attributes are required on this class, but could be provided as `""`. I'm fine dealing with that edge case since the user would effectively be choosing to break it. That becomes more of an issue when this winds up in a base class in `dbt-core` (down the road). If you're still concerned, I would address it with a validation rule, checking that all three parts are of non-zero length (rather than complicate this logic).

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: ditto to above

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: It get's used in `super().from_relation_results()`, which merely chains together this and `from_dict()`. This may not make as much sense here, but I'm going for a consistent look/feel across adapters and `RelationConfigBase` classes. When there are attributes which are themselves `RelationConfigBase` classes, e.g. `target_lag` below, then we need this method as a stopping point to produce a `dict`. Hence it gets called if this instance were an attribute on a bigger object.

tl;dr; Consistency? And to fit into the base class framework

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: I basically looked at what Snowflake returned on the most basic dynamic table. That being said, we control the format of what goes in there since we're writing the DDL that wraps this statement (the user's SQL). We're effectively unwrapping the statement here. I'm sure we'll find an edge case eventually and this will grow. The upside? We can actually compare the query of the object in the database and the query the user provided to see if there's a difference before deciding to deploy (this could be used to simplify view processing in the future).

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: It's used in the alter statement in the jinja here: https://github.com/dbt-labs/dbt-snowflake/blob/f9e80b679bf08360b59ee2c0349caeef465d7157/dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql#L24

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: If I remember correctly, `dataclass` yells at you if you have an `Optional` and you don't provide the variable because you haven't defined a default, even if that default is `None`. I think it doesn't assume. That being said, these are one of two varieties. There can either be a change that can only happen once (e.g. `target_lag`), or a set of changes all of the same type (e.g. `indexes` on Postgres). So I'm also being explicit that the null case here is `None` and not `set()`. I've also thought about collapsing this entirely and just having one single set of changes, but that leaves the parsing of "what _type_ of change is this?" to the caller, which in this case is jinja.

dbt/adapters/snowflake/relation_configs/target_lag.py
mikealfare: I think this is already guarding against the `0 hours` scenario because it's requiring a positive number. But it's a bit hidden since I do a check against the type and then combine it with another check. I'll break this into two checks: positive integer, and >= 60 if your unit is seconds.

dbt/adapters/snowflake/relation_configs/target_lag.py
mikealfare: It's a poor man's `mashumaro`, `pydantic`, `marshmallow`, etc. I could see replacing this in the long term with a framework like that. The idea is that there are several ways to create this "generic" dynamic view representation. This dataclass is just one giant set of parsers with validation rules. The two parsers I found so far is database > config and user-config > config. I wouldn't be surprised if we wind up with more. Maybe someone wants to get crazy and do a Redshift MV user-config > Snowflake DT config to do a migration. Whatever it is, this config class is our anchor. And we generally want two representations of this config. We want a nice shiny, python object with dot path-ing, etc. so that the jinja reads cleanly, other dependent code reads cleanly, etc. And then we usually want the dictionary version that is ubiquitous; it's nice for logs, it translates out of python nicely if necessary (basically json). In order to do that in an hierarchical object, each object needs both and we need to split apart the foreign format > dict and dict > object computation. A framework like `mashumaro` does all of that behind the scenes so you don't need to worry about it. However, we still have the added complexity of needing more than one parser since our objects come from more than one source. I'd be happy to discuss this further, I think it need sync discussions to make sense.

dbt/adapters/snowflake/relation_configs/target_lag.py
mikealfare: Yeah, that sounds right. I think I originally had this in an if:
```python
if target_lag := model_node.config.extra.get("target_lag"):
    try:
        duration, period = target_lag.split(" ")
    except IndexError:
        duration, period = None, None
```
But then I removed the if and forgot about that case. I'll update.

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
mikealfare: Yes, it's the former. It's the pieces that go into `dynamic_table/materialization.sql`. When the materialization is in `dbt-core`, this boils down to just a file `dynamic_table.sql`. I broke it out here to emphasize that the materialization is local (and because it's a lot of code with slightly different uses). Agreed, I should (and will) add a doc block at the top.

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
mikealfare: > Also, there is an argument a good in-code comment says why rather than what
Agreed. I'll update this.

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
mikealfare: Ah yes, I forgot about that. Nice catch. To be honest, I made it look like that because there were extra returns in the logs and this looked prettier, lol. Will change.

dbt/include/snowflake/macros/materializations/dynamic_table/ddl.sql
mikealfare: This was the recommended way to read the results from `show dynamic tables` according to the docs I found. I could see if I can find that again if you want. I'm not tied to it though; it worked so I went with it.

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
mikealfare: This is _very_ similar to the new built-in materialized view materialization in `dbt-core` (if you find/replace "dynamic table" with "materialized view", you might even say they're the same). That being said, I had these comments at the time to try and lead myself (and the next unfortunate soul) through this giant, nested, jinja-dwelling if block. But it goes back to your comment of "comments should say why, not what". And I think what this is suggesting is that maybe it should be its own macro, which would self-document what I'm doing here. I'll try some combination of all of that to make this better.

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
mikealfare: I shied away from a qualifier like `target_` or `existing_` when there were not conflicting versions of the same object. In hindsight, this should probably be generic, which would then look like `target_relation_name`, and leave the calculation of the name up to the caller; e.g. in Redshift MVs, this would be `materialized_view.mv_name` since that's how Redshift refers to that attribute in their docs.

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
mikealfare: Every other materialization in `dbt-snowflake` did this. And it seemed like a functionality we supported across the adapter; so I added it in.

dbt/adapters/snowflake/relation.py
VersusFacit: So what happens when this crashes on a user for some reason down the line against all our assumptions? We fine with this being the possible bug report?

dbt/adapters/snowflake/relation.py
VersusFacit: I'd love a `﻿﻿new_dynamic_table` name everywhere personally!

dbt/adapters/snowflake/relation.py
VersusFacit: Bruh I trust you but wow. Conventions be something. I'm happy with the compromise hehe

dbt/adapters/snowflake/relation.py
VersusFacit: I'm fine with more specific args and doc strings!

.changes/unreleased/Features-20230620-224944.yaml
VersusFacit: Oh? Interesting. Makes sense to leave as is

dbt/adapters/snowflake/relation_configs/dynamic_table.py
VersusFacit: I _love_ `fully_qualified_path`.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
VersusFacit: I don't love this but I'll take the explanation.

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
VersusFacit: Ah makes sense! Good catch.

dbt/adapters/snowflake/relation_configs/target_lag.py
VersusFacit: Oh excellent point about the boolean logic though!

dbt/include/snowflake/macros/materializations/dynamic_table/materialization.sql
VersusFacit: That'd be rad!

dbt/adapters/snowflake/relation.py
mikealfare: Yeah, I should probably wrap that in a `try` / `catch` and make it pretty, fair point.

Fleid:
> Refresh existing dynamic table with no changes

Hey @mikealfare looking at the original spec, the DT should not be refreshed if there are no changes and their status isn't `failed`.
I've re-worded the expected lifecycle [here](https://github.com/dbt-labs/dbt-snowflake/issues/667), hopefully clarifying it.

> full refresh on `warehouse` changes
I'm not sure what that means? Is that still an `ALTER` statement? Or are you doing a `CREATE OR REPLACE`?

mikealfare:
This PR is way out of date. The branch will be saved for reference, but the PR is stale.

PR #661, opened 2023-06-15
Try dbt-core#7368

jtcohen6:
Run CI with updated dependency pins


jtcohen6:
Tests are failing in a way that seems legitimately related to these changes. I don't have a chance to dig in right now, but it will take some digging.

mikealfare:
@jtcohen6, do you still need this branch? Or can we delete it?

jtcohen6:
Deleted!

PR #689, opened 2023-07-13
Try out adapter tests for query comments

dbeatty10:
Testing https://github.com/dbt-labs/dbt-core/pull/7928 against each adapter.

If all tests pass, then this can be closed without merging.


dbeatty10:
This has served its purpose and can be closed without further action.

PR #727, opened 2023-08-08
[ADAP-774] fix snowflake_warehouse issue with dynamic tables

McKnight-42:
resolves #
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

### Problem
users are being given incorrect information around required configs for `dynamic tables` in which they are being told that `warehouse` is not provided but they should actually be providing `snowflake_warehouse`.

### Solution
TODO:
- [ ] create changelog

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #732, opened 2023-08-09
[ADAP-631] Convert Target_lag to str type to add Downstream option

McKnight-42:
resolves #

### Problem
We need to allow for all target lags variations that Snowflake accepts. Variations include:

### Solution
change expected type to str

TODO:
- [x] merge ADAP-774
- [ ] update functional tests to check for `downstream` option.
- [ ] work on validation options (if possible)
- [x] create changelog

### Checklist


dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -107,7 +87,7 @@ def requires_full_refresh(self) -> bool:
 
 @dataclass
 class SnowflakeDynamicTableConfigChangeset:
-    target_lag: Optional[SnowflakeDynamicTableTargetLagConfigChange] = None
+    target_lag: Optional[SnowflakeDynamicTableWarehouseConfigChange] = None
```
mikealfare, line 90: This would be a change that looks like `SnowflakeDynamicTableWarehouseConfigChange`, but I would keep it separate, so `SnowflakeDynamicTableTargetLagConfigChange`. The class would be almost identical to the warehouse one. Leave it in this file, just like the warehouse one.

dbt/adapters/snowflake/relation.py
```diff
@@ -45,10 +44,7 @@ def dynamic_table_config_changeset(
         config_change_collection = SnowflakeDynamicTableConfigChangeset()
 
         if new_dynamic_table.target_lag != existing_dynamic_table.target_lag:
-            config_change_collection.target_lag = SnowflakeDynamicTableTargetLagConfigChange(
```
mikealfare, line 48: I think you need to keep this.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
McKnight-42: done

dbt/adapters/snowflake/relation.py
McKnight-42: yeah re added the dataclass, was originally part of the target_lag.py stuff that we ripped out.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -76,6 +76,15 @@ def parse_relation_results(cls, relation_results: RelationResults) -> dict:
         return config_dict
 
 
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class SnowflakeDynamicTableTargetLagConfigChange(RelationConfigChange):
```
mikealfare, line 80: With this class here, we should be able to delete `relation_configs/target_lag.py`

mikealfare (dismissed):

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -81,21 +69,22 @@ def parse_relation_results(cls, relation_results: RelationResults) -> dict:
...
-                        dynamic_table
-                    )
-                }
-            )
-
         return config_dict
 
 
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class SnowflakeDynamicTableTargetLagConfigChange(RelationConfigChange):
```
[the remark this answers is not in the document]
mikealfare: Nope, this will pass straight through. If we did want validation, we would put it on `SnowflakeDynamicTableConfig`. But for now, we're going without until we settle on exactly what values we're accepting.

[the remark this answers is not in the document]
McKnight-42: okay so we are technically allowing any string value to be passed for now alright will finish up tests and then this should be ready for merger

mikealfare (commented):
Let's add a test that verifies that we can pass "downstream" into `target_lag` and successfully create a dynamic table.

McKnight-42:
> Let's add a test that verifies that we can pass "downstream" into `target_lag` and successfully create a dynamic table.

Added a test following the current method of alter and seeing if a run is successful till we find a way to check the `show` command stuff via test suite. also would want to flag that might be good to be able to pass new param into methods like `change_config_via_alter` in end version so we don't have to duplicate it for every version we want.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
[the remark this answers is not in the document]
mikealfare: > okay so we are technically allowing any string value
Correct, we're relying on the Snowflake check for now. We'll come back around and put some validation in later, probably in 1.7.

.changes/unreleased/Fixes-20230810-163232.yaml
mikealfare, line 1: I think this got caught in here accidentally.

mikealfare (approved):
Looks good; I would just check that extra changie log.

.changes/unreleased/Fixes-20230810-163232.yaml
McKnight-42: yeah forgot one for the #727 so added it hear instead of trying to do a entire revert, and add since we don't tie the pr to the changelog anymore.

PR #747, opened 2023-08-14
[Backport] set target_lag to a str to allow for 'downstream' as accepted value

McKnight-42:
resolves https://github.com/dbt-labs/dbt-snowflake/pull/732

### Checklist


mikealfare:
@McKnight-42 It looks like this is failing tests in 1.6 somehow. I reran it to make sure and the same tests failed (dynamic tables). Maybe it has something to do with the fact that we have two things to backport and they conflict?

McKnight-42:
> @McKnight-42 It looks like this is failing tests in 1.6 somehow. I reran it to make sure and the same tests failed (dynamic tables). Maybe it has something to do with the fact that we have two things to backport and they conflict?

I believe it is the warehouse field since we would be testing against 1.6.latest which doesn't have that fix in it yet, that adding will get this pr moving.

nathaniel-may (approved):

PR #758, opened 2023-08-29
ADAP-865: Parameterize where clause, add option to supply list of relations

### LINKED ISSUE #801: [ADAP-943] [ADAP-928] [Feature] Implement relation filtering on get_catalog macro

### Describe the feature

Implement the new get_catalog_relations macro from https://github.com/dbt-labs/dbt-core/pull/8648 at the adapter level.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

mikealfare:
resolves #801 

### Problem

The current implementation of `get_catalog` only supports schema-level limitation. This can take a long time when schemas have a lot of objects in them.

### Solution

Provide a way to pass in a list of relations instead of schemas. The new `relations` parameter can be a list of relations, a list of schema relation objects, or a mix.

### Checklist


dbt/include/snowflake/macros/catalog.sql
```diff
@@ -1,72 +1,97 @@
...
-              'Approximate size of the table as reported by Snowflake' as "stats:bytes:description",
-              (bytes is not null) as "stats:bytes:include",
-
-              'Last Modified' as "stats:last_modified:label",
-              to_varchar(convert_timezone('UTC', last_altered), 'yyyy-mm-dd HH24:MI'||'UTC') as "stats:last_modified:value",
-              'The timestamp for last update/change' as "stats:last_modified:description",
-              (last_altered is not null and table_type='BASE TABLE') as "stats:last_modified:include"
-
-          from {{ information_schema }}.tables
-          where (
```
mikealfare, line 36: Parameterize this where clause with the new macro below.

dbt/include/snowflake/macros/catalog.sql
```diff
@@ -1,72 +1,97 @@
...
-              ordinal_position as "column_index",
-              data_type as "column_type",
-              comment as "column_comment"
+                'Row Count' as "stats:row_count:label",
+                row_count as "stats:row_count:value",
+                'An approximate count of rows in this table' as "stats:row_count:description",
+                (row_count is not null) as "stats:row_count:include",
 
-          from {{ information_schema }}.columns
-          where (
```
mikealfare, line 57: Parameterize this where clause with the new macro below.

dbt/include/snowflake/macros/catalog.sql
```diff
@@ -70,3 +62,36 @@
   {{ return(run_query(query)) }}
 
 {%- endmacro %}
+
+
+{% snowflake__get_catalog_where_clause(schemas, relations) %}
+
+    {% if schemas is not none %}
```
mikealfare, line 69: This is the original where clause for backwards compatibility.

dbt/include/snowflake/macros/catalog.sql
```diff
@@ -70,3 +62,36 @@
...
+        where (
+            {%- for schema in schemas -%}
+                upper("table_schema") = upper('{{ schema }}')
+                {%- if not loop.last %} or {% endif -%}
+            {%- endfor -%}
+        )
+
+    {% elif relations is not none %}
```
mikealfare, line 77: This is the new where clause that can take a set or BaseRelation instances. It assumes that `schema` is defined on all of them and `identifier` is defined on the instances that represent tables. This has the effect of allowing a mixture of tables and sets, if we want to support that. I condition based on the existence of `identifier`, which is a bit loose. There is no `is_schema` property on `BaseRelation` and I don't know that we should add one since `Schema` is not in `RelationType`; but it's another option.

mikealfare:
@peterallenwebb Here's a POC for the `get_catalog` update for applied state.

dev-requirements.txt
```diff
@@ -1,7 +1,7 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=tests/adapter
+git+https://github.com/dbt-labs/dbt-core.git@paw/get-catalog-relations#egg=dbt-core&subdirectory=core
```
mikealfare, line 3: Point this back to `main` before merging

.changes/unreleased/Features-20230829-152412.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Support limiting get_catalog by object name
+time: 2023-08-29T15:24:12.649104-04:00
+custom:
+  Author: mikealfare
+  Issue: "4997"
```
dbeatty10, line 6: Can we link to an issue in the dbt-snowflake repo so that the link in the changelog will work?

Otherwise, it will link to a page that doesn't exist.

dbeatty10: [Here](https://github.com/dbt-labs/dbt-snowflake/blob/1.6.latest/CHANGELOG.md#under-the-hood-1) is an example from the dbt-snowflake 1.6 changelog that doesn't work when linked to a dbt-core issue:

> test foreign key + unique constraint rendering ([#7417](https://github.com/dbt-labs/dbt-snowflake/issues/7417))

dbt/include/snowflake/macros/catalog.sql
```diff
@@ -1,72 +1,123 @@
...
+        data_type as "column_type",
+        comment as "column_comment"
+    from {{ information_schema }}.columns
+{%- endmacro %}
+
+
+{% macro snowflake__get_catalog_results_sql() -%}
+    select *
+    from tables
+    join columns using ("table_database", "table_schema", "table_name")
```
graciegoheen, line 90:
```suggestion
    inner join columns using ("table_database", "table_schema", "table_name")
```

From our [sql style guide](https://docs.getdbt.com/guides/best-practices/how-we-style/2-how-we-style-our-sql):
- 👭🏻 Be explicit about your join type (i.e. write inner join instead of join).

.changes/unreleased/Features-20230829-152412.yaml
mikealfare: Fair point. I'll update that to this PR number (which actually works because GH doesn't care if something is an issue or a PR).

dbt/include/snowflake/macros/catalog.sql
mikealfare: To be fair, this is the code that was there initially. It's listed as a change because I had to move it into its own macro. That being said, the use of `join` over `inner join` is fairly ubiquitous and comes down to a taste thing. I prefer fewer words in the join clause because it makes the stack of tables easier to skim; they line up better than with a longer join syntax.

dbt/include/snowflake/macros/catalog.sql
```diff
@@ -1,72 +1,123 @@
...
+        )
+        {{ snowflake__get_catalog_results_sql() }}
+    {%- endset -%}
+
+    {{ return(run_query(query)) }}
+
+{%- endmacro %}
+
+
+{% macro snowflake__get_catalog_relations(information_schema, relations) -%}
```
dataders, line 20: what's the difference between `snowflake__get_catalog_relations()` and `snowflake__get_catalog()`?

dataders:
@mikealfare do you predict that there will be relevant adapter zone tests here?

McKnight-42 (approved):

PR #770, opened 2023-09-13
bug/Dynamic-table-comment-on-syntax

### LINKED ISSUE #769: [ADAP-885] [Bug] Dynamic table alter syntax and comment syntax is wrong

### Current Behavior

DBT tries to comment on the model with a wrong syntax statement 

`comment on dynamic_table model_name AS`

Instead of 

`comment on dynamic table model_name AS`

Also, during the alter process

it tries to do 

`alter dynamic_table sandbox.procs.customer alter `

instead of 

`alter dynamic table sandbox.procs.customer alter `

### Expected Behavior

`comment on dynamic table model_name AS`

and 

`alter dynamic table sandbox.procs.customer alter `

### Steps To Reproduce

To be able to change the 'dynamic_table' keyword to 'dynamic table'

### Relevant log output

_No response_

### Environment

```markdown
- OS:Mac os
- Python: 3.8.16
- dbt-core: 1.6.2
- dbt-snowflake: 1.6.2
```

### Additional Context

_No response_

kaarthik108:
resolves #769
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) N/A

### Problem

This will fix the comment syntax for dynamic tables

from `comment on dynamic_table` to `comment on dynamic table`

### Solution

Just splitting the relation_type based on '_' only on dynamic tables

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -171,9 +171,10 @@
 {% endmacro %}
 
 {% macro snowflake__alter_relation_comment(relation, relation_comment) -%}
-    {%- set relation_type = relation.type -%}
-    {%- if relation_type == 'dynamic_table' -%}
+    {%- if relation.is_dynamic_table -%}
         {%- set relation_type = 'dynamic table' -%}
+    {%- else -%}
+        {%- set relation_type = relation.type -%}
     {%- endif -%}
     comment on {{ relation_type }} {{ relation }} IS $${{ relation_comment | replace('$', '[$]') }}$$;
 {% endmacro %}
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -171,7 +171,11 @@
 {% endmacro %}
 
 {% macro snowflake__alter_relation_comment(relation, relation_comment) -%}
-  comment on {{ relation.type }} {{ relation }} IS $${{ relation_comment | replace('$', '[$]') }}$$;
+    {%- set relation_type = relation.type -%}
+    {%- if relation_type == 'dynamic_table' -%}
+        {%- set relation_type = 'dynamic table' -%}
+    {%- endif -%}
+    comment on {{ relation_type }} {{ relation }} IS $${{ relation_comment | replace('$', '[$]') }}$$;
 {% endmacro %}
 
 
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -182,7 +182,12 @@
 
 {% macro snowflake__alter_column_comment(relation, column_dict) -%}
     {% set existing_columns = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
-    alter {{ relation.type }} {{ relation }} alter
+    {% if relation.is_dynamic_table -%}
+        {% set relation_type = "dynamic table" %}
+    {% else -%}
+        {% set relation_type = relation.type %}
+    {% endif %}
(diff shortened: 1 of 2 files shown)


kaarthik108:
https://github.com/dbt-labs/dbt-snowflake/issues/769

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -171,7 +171,11 @@
 {% endmacro %}
 
 {% macro snowflake__alter_relation_comment(relation, relation_comment) -%}
-  comment on {{ relation.type }} {{ relation }} IS $${{ relation_comment | replace('$', '[$]') }}$$;
+    {%- set relation_type = relation.type -%}
+    {%- if relation_type == 'dynamic_table' -%}
```
mikealfare, line 175: Minor nits:

To avoid the hard coded string, this could be:
```
{%- if relation.is_dynamic_table -%}
```
And I would move the initial "set relation_type" into an else statement in this if block.

What you have works, I just think this would be more explicit.

mikealfare:
Thanks for the PR @kaarthik108! I just had two nits, but otherwise I'm good with it as long as tests pass.

dbeatty10:
@mikealfare a curious question:

Would it simplify the code paths if [this](https://github.com/dbt-labs/dbt-snowflake/blob/cf04fd7eb667f5554d79f7e6c69f19dd06b34f55/dbt/adapters/snowflake/relation_configs/policies.py#L12
):

```python
 DynamicTable = "dynamic_table" 
```

is updated to this?

```python
 DynamicTable = "dynamic table" 
```

We can imagine a world in which "this one weird trick" irons out everything nicely, and all the if/elses go away ✨

We can also imagine a word in which it creates a bunch of wrinkles elsewhere that we don't want 🙅

Obviously up to you either way you want to go, but at least a possibility to consider.

mikealfare:
> @mikealfare a curious question:
> 
> Would it simplify the code paths if [this](https://github.com/dbt-labs/dbt-snowflake/blob/cf04fd7eb667f5554d79f7e6c69f19dd06b34f55/dbt/adapters/snowflake/relation_configs/policies.py#L12):
> 
> ```python
>  DynamicTable = "dynamic_table" 
> ```
> 
> is updated to this?
> 
> ```python
>  DynamicTable = "dynamic table" 
> ```
> 
> We can imagine a world in which "this one weird trick" irons out everything nicely, and all the if/elses go away ✨
> 
> We can also imagine a word in which it creates a bunch of wrinkles elsewhere that we don't want 🙅
> 
> Obviously up to you either way you want to go, but at least a possibility to consider.

That's a good question, and one we've discussed a bit due to issues like this one. dbt's interface almost everywhere (I say "almost" because I can't confirm it off the top of my head) is snake case. After discussing, we landed on keeping it snake case for consistency. I also wouldn't want the enum to serve multiple purposes (and I don't like that we currently do that in other places). It should solely focus on identifying the type of relation we're working with. The string representation of the enum also happens to also be the text that gets put into the jinja template most times, specifically for tables in views. However, that obviously doesn't work for dynamic tables and materialized views, which have spaces. It also doesn't even work for tables/views all the time. IIRC, you can't drop a view in postgres by running `drop view name_of_my_view`; it throws a syntax error. You have to run `drop table name_of_my_view`. So we can't generalize this to `drop {{ relation.type }} {{ relation }}`. It also can't be generalized to `drop table {{ relation }}` because this fails for materialized views. This also falls down once you get to more complicated statements, such as the `create` statement. We want this to be uniform across all statements for the adapters that we maintain. So while it does create a lot of boilerplate, copy/paste code, we feel this is a better, more consistent, more maintainable approach.

mikealfare (approved):

PR #773, opened 2023-09-14
Redact values from logs due 'duplicate key' error

### LINKED ISSUE #772: [ADAP-886] [Feature] Redact raw data from Duplicate Field Key Error

### Describe the feature

In https://github.com/dbt-labs/dbt-snowflake/pull/378, redaction was added for duplicate row DML.  We have recently discovered a second type of error — duplicate field key — which can leak sensitive data into logs.  Here's an example of what an error looks like:

```
Duplicate field key 'my raw string goes here'
```

I'd like to propose adding similar redacting to this use case.  I've also contact Snowflake to confirm whether there are other related errors we may need to handle.

Wrap dbt and redact the logs before they hit downstream alerting tools.

### Who will this benefit?

Anyone wanting to stop data leaking into their logs.

Absolutely

### Anything else?

_No response_

jaypeedevlin:
**🚨 This is as yet untested, looking to CI tests to validate my method**

resolves #772 
Docs N/A

Assuming this is merged, it would be amazing if this could be backported into 1.6.x so we can take advantage in our org without having to go through a minor version upgrade (which there's a lot of work involved given our scale)

### Problem

We are leaking raw data into logs in certain cases.

### Solution

Copies the method that exists, placing code into a loop to handle any other future cases that may arise.

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/connections.py
+++ b/dbt/adapters/snowflake/connections.py
@@ -48,7 +48,11 @@ from dbt.ui import line_wrap_message, warning_tag
 
 logger = AdapterLogger("Snowflake")
 _TOKEN_REQUEST_URL = "https://{}.snowflakecomputing.com/oauth/token-request"
-ROW_VALUE_REGEX = re.compile(r"Row Values: \[(.|\n)*\]")
+
+ERROR_REDACTION_PATTERNS = {
+    "Row Values: [redacted]": re.compile(r"Row Values: \[(.|\n)*\]"),
+    "Duplicate field key '[redacted]'": re.compile(r"Duplicate field key '(.|\n)*'")
+}
 
 
 @dataclass
@@ -280,13 +284,14 @@ class SnowflakeConnectionManager(SQLConnectionManager):
         try:
             yield
         except snowflake.connector.errors.ProgrammingError as e:
-            unscrubbed_msg = str(e)
+            msg = str(e)
 
             # A class of Snowflake errors -- such as a failure from attempting to merge
             # duplicate rows -- includes row values in the error message, i.e.
             # [12345, "col_a_value", "col_b_value", etc...]. We don't want to log potentially
             # sensitive user data.
-            msg = re.sub(ROW_VALUE_REGEX, "Row Values: [redacted]", unscrubbed_msg)
+            for replacement_message, regex_pattern in ERROR_REDACTION_PATTERNS.items():
+                msg = re.sub(regex_pattern, replacement_message, msg)
 
             logger.debug("Snowflake query id: {}".format(e.sfqid))
             logger.debug("Snowflake error: {}".format(msg))
--- a/tests/functional/duplicate_key_unlogged/test_duplicate_key_not_in_exceptions.py
+++ b/tests/functional/duplicate_key_unlogged/test_duplicate_key_not_in_exceptions.py
@@ -29,4 +29,4 @@ class TestRowValuesNotInExceptions:
     def test_row_values_were_scrubbed_from_duplicate_merge_exception(self, project):
         result = run_dbt(["run", "-s", "model"])
         assert len(result) == 1
-        assert "Duplicate field key [redacted]" in result[0].message
+        assert "Duplicate field key '[redacted]'" in result[0].message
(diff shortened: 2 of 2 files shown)


tests/functional/duplicate_key_unlogged/test_duplicate_key_not_in_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+
+class TestRowValuesNotInExceptions:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"model.sql": _MODELS__view}
+
+    def test_row_values_were_scrubbed_from_duplicate_merge_exception(self, project):
+        result = run_dbt(["run", "-s", "model"])
```
dbeatty10, line 30: If you expect a dbt command to yield a non-zero exit status (i.e. fail / not pass), then you need to do something like [this](https://github.com/dbt-labs/dbt-snowflake/blob/903ccea98d354a7eaa7d3dfd13679dc29e4d4597/tests/functional/adapter/test_concurrency.py#L8) or else pytest will yield an error:

```suggestion
        result = run_dbt(["run", "-s", "model"], expect_pass=False)
```

dbt/adapters/snowflake/connections.py
```diff
@@ -48,7 +48,11 @@
 
 logger = AdapterLogger("Snowflake")
 _TOKEN_REQUEST_URL = "https://{}.snowflakecomputing.com/oauth/token-request"
-ROW_VALUE_REGEX = re.compile(r"Row Values: \[(.|\n)*\]")
+
+ERROR_REDACTION_PATTERNS = {
+    "Row Values: [redacted]": re.compile(r"Row Values: \[(.|\n)*\]"),
+    "Duplicate field key '[redacted]'": re.compile(r"Duplicate field key '(.|\n)*'")
```
dbeatty10, line 54: To pass the automated code formatting check:

```suggestion
    "Duplicate field key '[redacted]'": re.compile(r"Duplicate field key '(.|\n)*'"),
```

dbeatty10 (commented):
@jaypeedevlin thanks for raising this PR 🤩 !

Three things to get this passing CI:
1. Code suggestion to expect `dbt run` to fail when it raises the error
1. Code suggestion to apply the expected code format
1. Add a changelog entry via `changie new`

Just let @dataders or myself if you want an assist with any of the above.

jaypeedevlin:
@dbeatty10 thanks for the review — I believe this should now be good to go.  Assuming the merge goes through if there's any possibility to get this backported to 1.6.x that would be a massive help for us!

dbt/adapters/snowflake/connections.py
```diff
@@ -48,7 +48,11 @@
 
 logger = AdapterLogger("Snowflake")
 _TOKEN_REQUEST_URL = "https://{}.snowflakecomputing.com/oauth/token-request"
-ROW_VALUE_REGEX = re.compile(r"Row Values: \[(.|\n)*\]")
+
+ERROR_REDACTION_PATTERNS = {
```
mikealfare, line 52: I like that this is more configurable. Have you considered swapping the keys and the values? I could see adding to this in the future with more patterns; but some of those patterns may have the same replacement message, which wouldn't be possible with this setup.

tests/functional/redact_log_values/test_duplicate_key_not_in_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TestDuplicateKeyNotInExceptions:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"model.sql": _MODELS__view}
+
+    def test_row_values_were_scrubbed_from_duplicate_merge_exception(self, project):
+        result = run_dbt(["run", "-s", "model"], expect_pass=False)
+        assert len(result) == 1
+        assert "Duplicate field key '[redacted]'" in result[0].message
```
mikealfare, line 32: I think we should also assert that the value we want to remove is not in the logs. The likelihood is low, but it would still be possible to have both the redacted and unredacted versions in the logs with this test.

mikealfare (commented):
Thanks for the PR @jaypeedevlin! I just had two suggestions, otherwise I like the approach.

colin-rogers-dbt (approved):

PR #777, opened 2023-09-17
add transient configs

### LINKED ISSUE #776: [ADAP-901] [Bug] Not possible to set `transient` config on Python models

### Current Behavior

1. It doesn't seem like it's possible to set the `transient` [config](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#transient-tables) on Python models.
2. It's also defaulting to non-transient.

### Expected Behavior

`transient` config should be respected by the Python model.

### Steps To Reproduce

1. Project setup:

```yml
# dbt_project.yml
name: my_dbt_project
profile: all
config-version: 2
version: 1.0

models:
  my_dbt_project:
    +materialized: table
    # +transient: true    # This doesn't work for snek.py.
```

```python
# models/snek.py
import pandas as pd
def model(dbt, session):
    dbt.config(transient=True)
    # dbt.config(transient="true")    # Neither does using a string in a model level config.
    return pd.DataFrame({"id": [1]})
```

2. Run:

```sh
$ dbt --debug run -s snek

22:03:38  Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'start', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x10a0b9820>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x10910dbb0>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x10ccf6130>]}
22:03:38  Running with dbt=1.6.2
22:03:38  running dbt with arguments {'printer_width': '80', 'indirect_selection': 'eager', 'log_cache_events': 'False', 'write_json': 'True', 'partial_parse': 'False', 'cache_selected_only': 'False', 'profiles_dir': '/Users/jeremy/.dbt', 'fail_fast': 'False', 'warn_error': 'None', 'log_path': '/Users/jeremy/src/dbt-basic/logs', 'debug': 'True', 'version_check': 'True', 'use_colors': 'True', 'use_experimental_parser': 'False', 'no_print': 'None', 'quiet': 'False', 'log_format': 'default', 'static_parser': 'True', 'introspect': 'True', 'invocation_command': 'dbt --debug run -s snek', 'target_path': 'None', 'warn_error_options': 'WarnErrorOptions(include=[], exclude=[])', 'send_anonymous_usage_stats': 'True'}
22:03:39  Sending event: {'category': 'dbt', 'action': 'project_id', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x10910dbb0>]}
22:03:39  Sending event: {'category': 'dbt', 'action': 'adapter_info', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ed883d0>]}
22:03:39  Registered adapter: snowflake=1.6.2
22:03:39  checksum: 8d66c46a7c5c0e5717f369273dbde9db64740684d9ba42e6442c1808d698ae9f, vars: {}, profile: , target: , version: 1.6.2
22:03:39  Partial parsing not enabled
22:03:40  Sending event: {'category': 'dbt', 'action': 'load_project', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ef569a0>]}
22:03:40  Sending event: {'category': 'dbt', 'action': 'resource_counts', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ef84160>]}
22:03:40  Found 1 model, 0 sources, 0 exposures, 0 metrics, 373 macros, 0 groups, 0 semantic models
22:03:40  Sending event: {'category': 'dbt', 'action': 'runnable_timing', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ef84eb0>]}
22:03:40  
22:03:40  Acquiring new snowflake connection 'master'
22:03:40  Acquiring new snowflake connection 'list_development'
22:03:40  Using snowflake connection "list_development"
22:03:40  On list_development: /* {"app": "dbt", "dbt_version": "1.6.2", "profile_name": "all", "target_name": "sf", "connection_name": "list_development"} */
show terse schemas in database development
    limit 10000
22:03:40  Opening a new connection, currently in state init
22:03:41  SQL status: SUCCESS 281 in 2.0 seconds
22:03:42  On list_development: Close
22:03:42  Re-using an available connection from the pool (formerly list_development, now create_development_dbt_jyeo)
22:03:42  Creating schema "database: "development"
schema: "dbt_jyeo"
"
22:03:42  Using snowflake connection "create_development_dbt_jyeo"
22:03:42  On create_development_dbt_jyeo: /* {"app": "dbt", "dbt_version": "1.6.2", "profile_name": "all", "target_name": "sf", "connection_name": "create_development_dbt_jyeo"} */
create schema if not exists development.dbt_jyeo
22:03:42  Opening a new connection, currently in state closed
22:03:43  SQL status: SUCCESS 1 in 1.0 seconds
22:03:43  On create_development_dbt_jyeo: Close
22:03:43  Re-using an available connection from the pool (formerly create_development_dbt_jyeo, now list_development_dbt_jyeo)
22:03:43  Using snowflake connection "list_development_dbt_jyeo"
22:03:43  On list_development_dbt_jyeo: /* {"app": "dbt", "dbt_version": "1.6.2", "profile_name": "all", "target_name": "sf", "connection_name": "list_development_dbt_jyeo"} */
show terse objects in development.dbt_jyeo limit 10000
22:03:43  Opening a new connection, currently in state closed
22:03:44  SQL status: SUCCESS 0 in 1.0 seconds
22:03:44  On list_development_dbt_jyeo: Close
22:03:45  Sending event: {'category': 'dbt', 'action': 'runnable_timing', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ed8c0d0>]}
22:03:45  Concurrency: 1 threads (target='sf')
22:03:45  
22:03:45  Began running node model.my_dbt_project.snek
22:03:45  1 of 1 START python table model dbt_jyeo.snek .................................. [RUN]
22:03:45  Re-using an available connection from the pool (formerly list_development_dbt_jyeo, now model.my_dbt_project.snek)
22:03:45  Began compiling node model.my_dbt_project.snek
22:03:45  Writing injected SQL for node "model.my_dbt_project.snek"
22:03:45  Timing info for model.my_dbt_project.snek (compile): 10:03:45.365803 => 10:03:45.408906
22:03:45  Began executing node model.my_dbt_project.snek
22:03:45  Writing runtime python for node "model.my_dbt_project.snek"
22:03:45  Using snowflake connection "model.my_dbt_project.snek"
22:03:45  On model.my_dbt_project.snek: /* {"app": "dbt", "dbt_version": "1.6.2", "profile_name": "all", "target_name": "sf", "node_id": "model.my_dbt_project.snek"} */
WITH snek__dbt_sp AS PROCEDURE ()

RETURNS STRING
LANGUAGE PYTHON
RUNTIME_VERSION = '3.8'
PACKAGES = ('snowflake-snowpark-python')

HANDLER = 'main'
EXECUTE AS CALLER
AS
$$

import sys
sys._xoptions['snowflake_partner_attribution'].append("dbtLabs_dbtPython")


  
    
# snowflake
import pandas as pd
def model(dbt, session):
    dbt.config(transient=True)
    return pd.DataFrame({"id": [1]})


# This part is user provided model code
# you will need to copy the next section to run the code
# COMMAND ----------
# this part is dbt logic for get ref work, do not modify

def ref(*args, **kwargs):
    refs = {}
    key = '.'.join(args)
    version = kwargs.get("v") or kwargs.get("version")
    if version:
        key += f".v{version}"
    dbt_load_df_function = kwargs.get("dbt_load_df_function")
    return dbt_load_df_function(refs[key])


def source(*args, dbt_load_df_function):
    sources = {}
    key = '.'.join(args)
    return dbt_load_df_function(sources[key])


config_dict = {}


class config:
    def __init__(self, *args, **kwargs):
        pass

    @staticmethod
    def get(key, default=None):
        return config_dict.get(key, default)

class this:
    """dbt.this() or dbt.this.identifier"""
    database = "development"
    schema = "dbt_jyeo"
    identifier = "snek"
    
    def __repr__(self):
        return 'development.dbt_jyeo.snek'


class dbtObj:
    def __init__(self, load_df_function) -> None:
        self.source = lambda *args: source(*args, dbt_load_df_function=load_df_function)
        self.ref = lambda *args, **kwargs: ref(*args, **kwargs, dbt_load_df_function=load_df_function)
        self.config = config
        self.this = this()
        self.is_incremental = False

# COMMAND ----------

# To run this in snowsight, you need to select entry point to be main
# And you may have to modify the return type to text to get the result back
# def main(session):
#     dbt = dbtObj(session.table)
#     df = model(dbt, session)
#     return df.collect()

# to run this in local notebook, you need to create a session following examples https://github.com/Snowflake-Labs/sfguide-getting-started-snowpark-python
# then you can do the following to run model
# dbt = dbtObj(session.table)
# df = model(dbt, session)


def materialize(session, df, target_relation):
    # make sure pandas exists
    import importlib.util
    package_name = 'pandas'
    if importlib.util.find_spec(package_name):
        import pandas
        if isinstance(df, pandas.core.frame.DataFrame):
          session.use_database(target_relation.database)
          session.use_schema(target_relation.schema)
          # session.write_pandas does not have overwrite function
          df = session.createDataFrame(df)
    
    df.write.mode("overwrite").save_as_table('development.dbt_jyeo.snek', create_temp_table=False)

def main(session):
    dbt = dbtObj(session.table)
    df = model(dbt, session)
    materialize(session, df, dbt.this)
    return "OK"

  
$$
CALL snek__dbt_sp();
22:03:45  Opening a new connection, currently in state closed
22:03:50  SQL status: SUCCESS 1 in 5.0 seconds
22:03:50  Timing info for model.my_dbt_project.snek (execute): 10:03:45.410550 => 10:03:50.161700
22:03:50  On model.my_dbt_project.snek: Close
22:03:50  Sending event: {'category': 'dbt', 'action': 'run_model', 'label': '1b3c04de-d16a-426d-a053-f62a87ef3402', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11ff4f130>]}
22:03:50  1 of 1 OK created python table model dbt_jyeo.snek ............................. [SUCCESS 1 in 5.26s]
22:03:50  Finished running node model.my_dbt_project.snek
22:03:50  Connection 'master' was properly closed.
22:03:50  Connection 'model.my_dbt_project.snek' was properly closed.
22:03:50  
22:03:50  Finished running 1 table model in 0 hours 0 minutes and 10.20 seconds (10.20s).
22:03:50  Command end result
22:03:50  
22:03:50  Completed successfully
22:03:50  
22:03:50  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
22:03:50  Command `dbt run` succeeded at 10:03:50.642741 after 12.50 seconds
22:03:50  Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'end', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x10a0b9820>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11eed7340>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x11eda0250>]}
22:03:50  Flushing usage events
```

3. Check table type in Snowflake:

image

^ Kind is `TABLE` instead of `TRANSIENT`.

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.9.13
- dbt-core: 1.6.2
- dbt-snowflake: 1.6.2
```

### Additional Context

Looks like `save_as_table()` does support `transient` type (https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrameWriter.save_as_table)

https://github.com/dbt-labs/dbt-snowflake/blob/eb4fd78ce2e4200a7c437c6411b95c30a34b1840/dbt/include/snowflake/macros/materializations/table.sql#L55

I think we just need to get the config and then inject into the python code.

jeremyyeo:
resolves #776 

### Problem

Python models don't respect `transient` config.

### Solution

A snowpark table created via `save_as_table` can be transient via setting the param `table_type = 'transient'` (https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrameWriter.save_as_table).

1. Pull out the transient config setting in the CTAS (`snowflake__create_table_as()`) from being exclusive to SQL models.
2. Now that that config is available for Python models too - use that to determine the right value to set for the `table_type` arg.

There's probably something else to address.

1. Assuming users don't set `+transient` config anywhere (i.e. profile, project, model level).
2. By default, SQL tables are created as transient.
3. By default, Python tables are created as non-transient.

This change would make Python tables also be created as transient - in line with (2).

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/materializations/table.sql
+++ b/dbt/include/snowflake/macros/materializations/table.sql
@@ -38,7 +38,7 @@
 
 {% endmaterialization %}
 
-{% macro py_write_table(compiled_code, target_relation, temporary=False) %}
+{% macro py_write_table(compiled_code, target_relation, temporary=False, table_type='transient') %}
 {{ compiled_code }}
 def materialize(session, df, target_relation):
     # make sure pandas exists
@@ -52,7 +52,7 @@ def materialize(session, df, target_relation):
           # session.write_pandas does not have overwrite function
           df = session.createDataFrame(df)
     {% set target_relation_name = resolve_model_name(target_relation) %}
-    df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}})
+    df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}}, table_type='{{table_type}}')
 
 def main(session):
     dbt = dbtObj(session.table)
--- a/dbt/include/snowflake/macros/relations/table/create.sql
+++ b/dbt/include/snowflake/macros/relations/table/create.sql
@@ -1,6 +1,6 @@
 {% macro snowflake__create_table_as(temporary, relation, compiled_code, language='sql') -%}
+  {%- set transient = config.get('transient', default=true) -%}
   {%- if language == 'sql' -%}
-    {%- set transient = config.get('transient', default=true) -%}
     {%- set cluster_by_keys = config.get('cluster_by', default=none) -%}
     {%- set enable_automatic_clustering = config.get('automatic_clustering', default=false) -%}
     {%- set copy_grants = config.get('copy_grants', default=false) -%}
@@ -46,7 +46,8 @@
       {%- endif -%}
 
   {%- elif language == 'python' -%}
-    {{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary) }}
+    {%- set table_type = 'transient' if transient else '' -%}
+    {{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary, table_type=table_type) }}
   {%- else -%}
       {% do exceptions.raise_compiler_error("snowflake__create_table_as macro didn't get supported language, it got %s" % language) %}
   {%- endif -%}
--- a/.changes/unreleased/Fixes-20230918-105721.yaml
+++ b/.changes/unreleased/Fixes-20230918-105721.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Make python models use transient config
+time: 2023-09-18T10:57:21.113134+12:00
+custom:
+  Author: jeremyyeo
+  Issue: "776"


dbeatty10:
### User experience change

> There's probably something else to address.
> 
> 1. Assuming users don't set +transient config anywhere (i.e. profile, project, model level).
> 1. By default, SQL tables are created as transient.
> 1. By default, Python tables are created as non-transient.
> 
> This change would make Python tables also be created as transient - in line with (2).

Changing the default behavior for Python tables to be created as `transient` instead of non-transient seems reasonable to me since it would bring it in-line with the default value for SQL tables. It makes sense for them to have the same default.

#### The difference

The key difference is that [transient](https://docs.snowflake.com/en/user-guide/tables-temp-transient#transient-tables) tables do not have a Fail-safe period (whereas permanent tables do). So a user can trade data protection and recovery for lower cost.

#### The consequences

If this change is adopted as-is, then users will need to explicitly add `transient: false` / `dbt.config(transient=False)` if they want to keep the Fail-safe period's data protection and recovery abilities (along the associated costs).

### Documentation

It would be nice to add some documentation in docs.getdbt.com how to configure the `transient` boolean (including explanation of the default value).

I don't know if [this](https://docs.getdbt.com/docs/build/python-models#specific-data-platforms) is the best place or if there's a better place.

We'd probably want to show the non-default config:

1. Project setup:

```yml
# dbt_project.yml
name: my_dbt_project
profile: all
config-version: 2
version: 1.0

models:
  my_dbt_project:
    +materialized: table
    +transient: false
```

```python
# models/snek.py
import pandas as pd
def model(dbt, session):
    dbt.config(transient=False)
    return pd.DataFrame({"id": [1]})
```

### Testing

We'd probably want to add some test case to make sure that both transient true & false behave as intended.

Would probably be accomplished by using config like shown above and then checking the value in the information schema afterwards.

Would you be able to add those tests @jeremyyeo ? Or would you be looking for further guidance or assistance from the Adapters team on the testing portion?

jeremyyeo:
> Would probably be accomplished by using config like shown above and then checking the value in the information schema afterwards.
> Would you be able to add those tests @jeremyyeo ? Or would you be looking for further guidance or assistance from the Adapters team on the testing portion?

Certainly would love some guidance added those - couldn't find any examples in the repo that does this (but could just be me essentially having a lack of testing knowledge).

dbeatty10:
> Certainly would love some guidance added those - couldn't find any examples in the repo that does this (but could just be me essentially having a lack of testing knowledge).

@jeremyyeo We'll need to get insight from someone on the Adapters team of what kind of tests they'd like for this.

I didn't see any tests here that seemed like a perfect match:
- GitHub search: [`repo:dbt-labs/dbt-snowflake transient`](https://github.com/search?q=repo%3Adbt-labs%2Fdbt-snowflake%20transient&type=code)

colin-rogers-dbt:
@jeremyyeo this is a great change

To add some tests:
1. Create an example python model as a string like [here](https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/functional/adapter/test_python_model.py#L8). You should be able to just copy the `models__simple_python_model` fixture and add the transient config to it
2. Add a new test class like: 
```
class TestTransientModelSnowflake:
    def models(self):
        return {"transient_python_model.py": models__transient_python_model}

    def test_transient_model(self, project):
        run_dbt(["run"])
 ```

File: dbt/include/snowflake/macros/materializations/table.sql [hunk 1]
```diff
@@ -38,7 +38,7 @@
 
 {% endmaterialization %}
 
-{% macro py_write_table(compiled_code, target_relation, temporary=False) %}
+{% macro py_write_table(compiled_code, target_relation, temporary=False, table_type='transient') %}
```
mikealfare, line 41: I think I'd rather pass `transient` as a boolean into this macro and then handle the translation here for the call to `df.write.mode()`; the translation is specific to this call and we assume this is a boolean in the user config. Right now we have the translation in the `create` macro, which maps to `"transient"` or `""`. Also, is empty string correct here? What happens if we submit a call to `df.write.mode()` with `table_type=""` (versus omitting the kwarg `table_type` entirely)?

dbeatty10: > Also, is empty string correct here? What happens if we submit a call to `df.write.mode()` with `table_type=""` (versus omitting the kwarg `table_type` entirely)?

@mikealfare that's a wise question 🧠

I didn't test this out personally, but the comments in the snowpark source code describe that case [here](https://github.com/snowflakedb/snowpark-python/blob/100071a27051aa15b1dad98900947e1a5119bbb9/src/snowflake/snowpark/dataframe_writer.py#L146-L148).

So the way it is implemented in this PR currently, the string value of `table_type` just becomes a pass-through in line 55 below:

```python
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}}, table_type='{{table_type}}')
```

jeremyyeo: From [Snowflake docs](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrameWriter.save_as_table) (and what Doug linked to):

> table_type – The table type of table to be created. The supported values are: temp, temporary, and transient. An empty string means to create a permanent table. Learn more about table types [here](https://docs.snowflake.com/en/user-guide/tables-temp-transient.html).

So effectively:

```py
# Table is transient.
save_as_table(..., table_type='transient')

# Table is non-transient.
save_as_table(..., table_type='')

# Table is non-transient (so default of `table_type` seems to be empty string).
save_as_table(...)
```

Could swap it out to (according to Mike's suggestion):

```sql
-- dbt/include/snowflake/macros/materializations/table.sql
{% macro py_write_table(compiled_code, target_relation, temporary=False, transient=False) %}
...
{% set table_type = 'transient' if transient else '' %}
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}}, table_type='{{table_type}}')
```

Not too sure if it's worth writing additional logic to omit (or not) the `table_type` arg when it is an empty string (or not) - i.e.

```sql
-- dbt/include/snowflake/macros/materializations/table.sql
{% macro py_write_table(compiled_code, target_relation, temporary=False, transient=False) %}
...
{% if transient %}
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}}, table_type='transient')
{% else %}
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}})
```

Happy to follow a suggested pattern @mikealfare

dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -46,7 +46,8 @@
       {%- endif -%}
 
   {%- elif language == 'python' -%}
-    {{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary) }}
+    {%- set table_type = 'transient' if transient else '' -%}
+    {{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary, table_type=table_type) }}
```
jeremyyeo, line 50: Change
```sql
{%- set table_type = 'transient' if transient else '' -%}
{{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary, table_type=table_type) }}
```

to
```sql
{{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary, transient=transient) }}
```

and handle within `py_write_table` instead if we want to proceed with Mike's suggestion above (https://github.com/dbt-labs/dbt-snowflake/pull/777/files#r1330527575).

dbt/include/snowflake/macros/materializations/table.sql
jeremyyeo: Ftr, the docs linked above point out that the kwarg `create_temp_table` is deprecated:

> create_temp_table – (Deprecated) The to-be-created table will be temporary if this is set to True.

Which of course we still use in our `save_as_table()` method calls - so potentially would be a good time to replace that with `table_type` kwarg entirely.

Go from (status quo):
```python
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}})
```

Or (this PR's original intended change):

```python
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', create_temp_table={{temporary}}, table_type='{{table_type}}')
```

To just:
```python
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='{{table_type}}')
```

And then - we'll have all types via:

```python
# Table is transient.
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='transient')

# Table is non-transient.
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='')

# Table is temporary.
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='temporary')
```

mikealfare: Sorry for the delay, just getting around to plowing through my GH notifications. All of this context helps a lot.

Given that SF is deprecating `create_temp_table` in favor of `table_type`, I agree with using `table_type` in the signature, as a string. And I would default it to empty string in alignment with the docs. This also keeps config parsing out of a macro that otherwise does not care about the jinja global `config` that's floating around.

I wouldn't worry about removing the `table_type` argument if `table_type` is not in the call. However, I think we need to be cognizant of backwards compatibility for the `temporary` argument in the `py_write_table` macro. So that means we need to take both arguments in the `py_write_table` macro and combine them. @dbeatty10, correct me if I'm wrong here, but I think that amounts to something like this:

```
{% if temporary == True %}
{% set table_type = "temporary" %}  -- hence override the value of `table_type` that was passed in
{% else %}
-- this else is not needed, but communicates the concept
-- keep the value of `table_type` that was passed in (which could be the default empty string)
{% endif %}
```
Then we update the call to `save_as_table` to exclude `create_temp_table` in alignment with your third option above:
```python
df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='{{table_type}}')
```

dbt/include/snowflake/macros/relations/table/create.sql
mikealfare: I would do a few things here in alignment with my comment above. I would update the definition of `table_type` here to consider `temporary` as discussed above. And then I would not pass `temporary` into the `py_write_table` call at all. That tidies up all the config parsing. You would still need to keep that if block in the `py_write_table` macro in the even that other folks are using it; but then when we eventually retire the `temporary` argument in that macro, we don't need to remember to come back here and deal with it here.

dbt/include/snowflake/macros/materializations/table.sql
dbeatty10: That sounds about right @mikealfare.

### Logic

To support both backwards-compatibility as well as forward-facing use cases, I'd suggest that the new `table_type` parameter takes precedence over `temporary` (but only when it is specified).

So like this (in alignment with the Snowflake docs [here](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/latest/api/snowflake.snowpark.DataFrameWriter.save_as_table#:~:text=create_temp_table%20%E2%80%93%20(Deprecated)%20The,types%20here.)):
- use `table_type` when specified
- use `"temporary"` as the table type only when `temporary` is `True` and `table_type` is not specified
- default the table type to `"transient"` if all else fails

e.g., something similar to what you wrote, but with the precedence flipped:

### An untested implementation

```sql
{% macro py_write_table(compiled_code, target_relation, temporary=False, table_type=none) %}

...

{% if table_type is none and temporary %}
    {% set table_type = "temporary" %}
{% elif table_type is none %}
    {# Default to "transient", just like dbt SQL tables in Snowflake #}
    {% set table_type = "transient" %}
{% elif table_type == "permanent" %}
    {# Snowflake treats "" as meaning "permanent" #}
    {% set table_type = "" %}
{% endif %}

...

df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='{{table_type}}')
```

### Unnecessarily complicated?

This code might appear unnecessarily complicated on first blush. But it's a four-fold consequence of:

1. Goal of giving precedence to `table_type` when a consumer uses the 3-parameter version of the `py_write_table` macro.
1. Goal of being backwards-compatible when a consumer uses the 2-parameter version of the `py_write_table` macro.
1. Goal of using `"transient"` as the default value (unless overridden)
1. Snowflake docs say: "The supported values of `table_type` are: `temp`, `temporary`, and `transient`. An empty string means to create a permanent table."

mikealfare: I'm fine with this, but I have one suggestion.

Since we're mimicking the behavior of `py_write_table` and we want folks to use `table_type` moving forward, we want to support `table_type="transient"`, `table_type="temp"`, and `table_type="temporary"`. This logic does actually do that, but I read it multiple times before I realized that happens because we're updating the parameter that comes in, hence it passes straight through. For the sake of readability, especially because it's jinja, I'd like to add an else clause to the if block that just contains the comment "otherwise leave table_type as it is" (or something along those lines). It would save folks some time in the future.

> {# Snowflake treats "" as meaning "permanent" #}
I completely misread the docs. I conflated this with the default value of "transient". It's kind of wild that empty string is a valid value that means something and is also not the default.

dbt/include/snowflake/macros/relations/table/create.sql
mikealfare: @dbeatty10 I think this still aligns with what you're saying above, correct? We're effectively forcing the use of `table_type` here.

dbt/include/snowflake/macros/materializations/table.sql
dbeatty10: Wild indeed! 🤠

Your suggestion works for me 👍

Here's what it would look like after adding that piece:

```python
{% macro py_write_table(compiled_code, target_relation, temporary=False, table_type=none) %}

...

{% if table_type is none and temporary %}
    {% set table_type = "temporary" %}
{% elif table_type is none %}
    {#- Default to "transient", just like dbt SQL tables in Snowflake -#}
    {% set table_type = "transient" %}
{% elif table_type == "permanent" %}
    {#- Snowflake treats "" as meaning "permanent" -#}
    {% set table_type = "" %}
{% else %}
    {#- Otherwise leave table_type as it is -#}
{% endif %}

...

df.write.mode("overwrite").save_as_table('{{ target_relation_name }}', table_type='{{table_type}}')
```

dbt/include/snowflake/macros/relations/table/create.sql
dbeatty10: @mikealfare there's one more thing to handle here:
- 👉 dbt-snowflake has a boolean config named [`transient `](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-transient-tables-in-dbt_projectyml) that [defaults to `None`](https://github.com/dbt-labs/dbt-snowflake/blob/273694b38ed05d046a6b284a3ac396478a241bfa/dbt/adapters/snowflake/impl.py#L25)

### Which brings us to three-valued logic ...

There's three-valued logic to handle for the `transient` boolean config:
- `None` (pass-through and let `py_write_table` decide what to do)
- `True` (use `transient`)
- `False` (use `permanent`?)

### How 'bout this?

So I think we'd need handle the null case first to be sure the `table_type` is set (or not set!) correctly:

```sql
{% if transient is none %}
    {% set table_type = none %}
{% elif transient %}
    {% set table_type = "transient" %}
{% else %}
    {% set table_type = "permanent" %}
{% endif %}

{{ py_write_table(compiled_code=compiled_code, target_relation=relation, table_type=table_type) }}
```

This is assuming an implementation of `py_write_table` like [our most recent iteration](https://github.com/dbt-labs/dbt-snowflake/pull/777#discussion_r1353680684).

mikealfare: If `transient` is `None`, then wouldn't we set `table_table` based on `temporary` since we're not passing `temporary` into `py_write_table` anymore? Put another way, I don't think you would be able to create a table with `table_type="temporary"` in your logic flow without also passing `temporary` into the call to `py_write_table`. And I think we don't want to do that from what we said.

```
{% if transient is none and temporary %}
    {% set table_type = "temporary" %}
{% elif transient is none %}
    {% set table_type = none %}
{% elif transient %}
    {% set table_type = "transient" %}
{% else %}
    {% set table_type = "permanent" %}
{% endif %}
```

dbeatty10: So we can assess all the proposed logic with full context, I'm going to raise an alternative PR that is branched off of this one.

Should be able to do that sometime within the next 2 hours.

Then we can hammer on it to make sure all the corner cases are covered.

dbeatty10: @mikealfare and @jeremyyeo

[Here](https://github.com/dbt-labs/dbt-snowflake/compare/0bf3822d90d7c64b15270e5e510861786562a92c...7074d92cebc2f5d8b435b5bf2f88e9d260f6cdfe) is some **untested**(!) draft code that I think is backwards-compatible if anyone is relying on the original signature:
```sql
py_write_table(compiled_code, target_relation, temporary=False)
```

That draft code can be simplified quite a bit if we choose to make an breaking/incompatible change of the signature to be this instead:
```sql
py_write_table(compiled_code, target_relation, table_type="")
```

mikealfare: The `table.sql` file in your draft code looks right. But the `create.sql` file doesn't:

```
{% if temporary -%}
  {%- set table_type = "temporary"
{%- elif transient -%}
  {%- set table_type = "transient"
{%- else -%}
  {%- set table_type = ""
{%- endif %}
```

If I'm following, we said that we want the use of `table_type`, which I think aligns with the use of `transient`, to override `temporary`. However, in the code block it looks like `temporary` takes effect over `transient`. Is that right?

dbeatty10: Yes, sharp eyes -- you read that right!

After examining the [logic for `sql` tables](https://github.com/dbt-labs/dbt-snowflake/blob/ff08a1cd9ea61bb5e54441a4a84d507ca461eedd/dbt/include/snowflake/macros/relations/table/create.sql#L20-L24) in dbt-snowflake it became clear that it prioritizes the `temporary` config over the `transient` config.

So I flipped-flopped from our earlier discussions and switched the implementation to standardize on the behavior of `sql` tables.

Specifically, I just re-factored the code so that [this logic](https://github.com/dbt-labs/dbt-snowflake/blob/ff08a1cd9ea61bb5e54441a4a84d507ca461eedd/dbt/include/snowflake/macros/relations/table/create.sql#L20-L24) applies to **both** `sql` and `python` tables.

Here is the relevant portion of the diff:

image

dbeatty10: @mikealfare Currently failing CI, but draft PR is up here: https://github.com/dbt-labs/dbt-snowflake/pull/802

mikealfare: I put a comment there. You're missing the close braces on the sets, so the code quality failed. Will we be moving forward with 802 then as the primary PR?

dbeatty10: Yes, let's consider https://github.com/dbt-labs/dbt-snowflake/pull/802 the primary PR. It preserves @jeremyyeo's commits history and authorship credit.

dbeatty10:
Closing in favor of https://github.com/dbt-labs/dbt-snowflake/pull/802

Thanks for all your work on this @jeremyyeo between finding the issue, explaining it so thoroughly (including links to Snowflake docs), and even fixing it! 🏆

Your authorship and commits are preserved within #802 -- @mikealfare and I just added some tests and tweaked the logic slightly.

PR #796, opened 2023-10-09
Implementation of metadata-based freshness

### LINKED ISSUE #785: [ADAP-910] Support Metadata Freshness
### Describe the feature

Support metadata-based freshness by implementing the new macro and feature flag described in [#8704](https://github.com/dbt-labs/dbt-core/issues/8704).

### Who will this benefit?

Everyone who wants faster freshness results from Snowflake!

peterallenwebb:
resolves #785

### Checklist


dbt/include/snowflake/macros/metadata.sql
```diff
@@ -0,0 +1,19 @@
+{% macro snowflake__get_relation_last_modified(information_schema, relations) -%}
+
+  {%- call statement('last_modified', fetch_result=True) -%}
+        select table_schema as schema,
+               table_name as identifier,
+               last_altered as last_modified,
+               {{ current_timestamp() }} as snapshotted_at
+        from {{ information_schema }}.tables
+        where (
```
peterallenwebb, line 9: Should we check whether is_active is true?

dev-requirements.txt
```diff
@@ -1,6 +1,6 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt-core.git@paw%2Fmetadata-freshness#egg=dbt-core&subdirectory=core
```
McKnight-42, line 3: should we change `%2` to a `/`? don't recall having to make this swap

dbt/adapters/snowflake/impl.py
```diff
@@ -259,3 +276,15 @@ def valid_incremental_strategies(self):
     def debug_query(self):
         """Override for DebugTask method"""
         self.execute("select 1 as id")
+
+    def capabilities(self) -> List[CapabilitySupport]:
```
mikealfare, line 280: How would this be used? My gut is that the value of this is for `dbt-core` to inspect the adapter for a given capability, regardless of what's in the adapter. That would mean that this needs to be the full enum, not just what's in the adapter.

tests/functional/adapter/test_get_last_relation_modified.py
```diff
@@ -0,0 +1,46 @@
...
+    freshness:
+      warn_after: {count: 10, period: hour}
+      error_after: {count: 1, period: day}
+    schema: "{{ env_var('DBT_TEST_SCHEMA_NAME_VARIABLE') }}"
+    tables:
+      - name: test_table
+"""
+
+
+class TestGetLastRelationModified:
```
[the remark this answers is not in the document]
peterallenwebb: Yeah. That is the issue. We would need a better mechanism for mocking out the warehouse itself, at least partially. It's something to consider in the long run, but I've already bitten off a lot with this changeset.

dev-requirements.txt
mikealfare: You mean you don't manually type it in and you copy/pasta from the URL? Blasphemy!

dbt/include/snowflake/macros/metadata.sql
```diff
@@ -0,0 +1,19 @@
+{% macro snowflake__get_relation_last_modified(information_schema, relations) -%}
+
+  {%- call statement('last_modified', fetch_result=True) -%}
+        select table_schema as schema,
```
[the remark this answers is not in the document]
peterallenwebb: I'm following the pattern set out by get_catalog here, but I had the same thought.

It's not clear to me what the true best practice would be. We could prepend the database name to information schema if there is the possibility that the database implied by the active connection would not match the database of the information schema we are interested in.

dbt/include/snowflake/macros/metadata.sql
```diff
@@ -0,0 +1,19 @@
...
+               table_name as identifier,
+               last_altered as last_modified,
+               {{ current_timestamp() }} as snapshotted_at
+        from {{ information_schema }}.tables
+        where (
+          {%- for relation in relations -%}
```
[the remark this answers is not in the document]
peterallenwebb: Yes. That is definitely the assumption here. It will definitely be true for the current use case on the core side.

dbt/adapters/snowflake/impl.py
```diff
@@ -259,3 +276,15 @@ def valid_incremental_strategies(self):
...
+        return [c for c in self._capabilities.values()]
+
+    def capability_support(self, capability: Capability) -> CapabilitySupport:
```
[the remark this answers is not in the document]
peterallenwebb: Done

dbt/adapters/snowflake/impl.py
peterallenwebb: I believe the situation is clarified with the new design, where the failure of an adapter to declare support results in the state of support simply being "Unknown" from core's perspective.

dbt/include/snowflake/macros/metadata.sql
peterallenwebb: @aranke It turns out this particular table does not include dropped tables.

dbt/include/snowflake/macros/metadata.sql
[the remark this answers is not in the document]
mikealfare: If that was the pattern, I'd stick with it. It's probably that it's a database-specific information schema.

mikealfare (approved):

PR #839, opened 2023-11-21
first pass: unit test typing

jtcohen6:
requires case-insensitivity changes in https://github.com/dbt-labs/dbt-core/pull/9131

### DIFF
--- a/dbt/adapters/snowflake/column.py
+++ b/dbt/adapters/snowflake/column.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Dict, ClassVar
 
 from dbt.adapters.base.column import Column
 from dbt.exceptions import DbtRuntimeError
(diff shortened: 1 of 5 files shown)


dbt/adapters/snowflake/column.py
```diff
@@ -1,11 +1,16 @@
 from dataclasses import dataclass
+from typing import Dict, ClassVar
 
 from dbt.adapters.base.column import Column
 from dbt.exceptions import DbtRuntimeError
 
 
 @dataclass
 class SnowflakeColumn(Column):
+    TYPE_LABELS: ClassVar[Dict[str, str]] = {
```
jtcohen6, line 10: The parent/default `Column.TYPE_LABELS` has `STRING -> TEXT`. I haven't included it here because these are true aliases on Snowflake.

PR #884, opened 2024-01-11
Add materialization comment in create statement

### LINKED ISSUE #883: [ADAP-1094] [Feature] Improving the addition of table comment/description in only one query

### Describe the feature

When adding a description in Snowflake dbt with persist_docs option, I want dbt to add the table comment as "CREATE OR REPLACE <view_name> COMMENT='<comment>' " statement, rather than executing a second SQL request to add the comment after the table creation.

_No response_

### Who will this benefit?

It will benefit dbt users using the dbt documentation as code feature to create the tables comments. The execution time will be quicker as only one request will be done instead of 2. Also, less rights would be needed as having a second request doing a "COMMENT" is asking for "ALTER" rights.

yes

### Anything else?

_No response_

pol-defont-reaulx:
resolves #883
[docs] N/A

### Problem

When adding a description in Snowflake dbt with persist_docs option, dbt is creating the table/view and then is doing other requests to add the table/view comment. It should add the table/view comment as "CREATE OR REPLACE <view_name> COMMENT='' " statement.

### Solution

I added a condition to add the "COMMENT" statement in the `macros/relations/<table or view>/create.sql` if persist_column_docs is set in the config and if the materialization description is not empty.
I also deleted the call of `persist_docs()` for relation for each concerned materialization to avoid the execution of the `COMMENT ON <table/view> <name> IS ...` query that was adding the comment before.

I think it could be improve by creating the columns comments also in the same query rather than in a 2nd.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240112-020600.yaml
+++ b/.changes/unreleased/Features-20240112-020600.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add table/view comment to the CREATE statement to execute less queries
+time: 2024-01-12T02:06:00.777689+01:00
+custom:
+  Author: pol-defont-reaulx
+  Issue: "883"
--- a/dbt/include/snowflake/macros/materializations/incremental.sql
+++ b/dbt/include/snowflake/macros/materializations/incremental.sql
@@ -136,7 +136,7 @@
    should_revoke(existing_relation.is_table, full_refresh_mode) %}
   {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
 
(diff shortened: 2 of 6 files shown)


nathaniel-may:
Hi @pol-defont-reaulx, thank you for filing this issue and putting in the work to open this PR. We went over it with the team, and there are several edge cases for other places we pull comments from that we would want to add tests for to make sure this doesn't affect that functionality as well. We suspect that these tests could be particularly tricky, and given our current work load we likely won't have the proper time to address this.

We appreciate your work here, but I am going to close the issue for now due to time constraints on our end.

PR #931, opened 2024-03-18
Add Snowflake internal build process workflow

VersusFacit:
resolves 38 of epic

This PR adds a workflow file for bigquery testing. It is based on https://github.com/dbt-labs/dbt-redshift/pull/732. I assume there will be some quirks to work out. Also, we'll need to add aws secrets.

### DIFF
--- a/.changes/unreleased/Features-20240318-033136.yaml
+++ b/.changes/unreleased/Features-20240318-033136.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: ' Add new workflow for internal patch releases'
+time: 2024-03-18T03:31:36.425644-07:00
+custom:
+  Author: versusfacit
+  Issue: "38"
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -1,0 +1,64 @@
+name: Release internal patch
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_number:
+        description: "The release version number (i.e. 1.0.0b1)"
+        type: string
+        required: true
+      sha:
+        description: "The sha to use (leave empty to use latest on main)"
+        type: string
+        required: false
+      package_test_command:
+        description: "Package test command"
+        type: string
+        default: "python -c \"import dbt.adapters.snowflake\""
+        required: true
+      dbms_name:
+        description: "The name of the warehouse the adapter connects to."
+        type: string
+        default: "snowflake"
+        required: true
+  workflow_call:
+    inputs:
+      version_number:
+        description: "The release version number (i.e. 1.0.0b1)"
+        type: string
+        required: true
+      sha:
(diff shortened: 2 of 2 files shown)


emmyoop (changes requested):
All the same questions/changes from https://github.com/dbt-labs/dbt-redshift/pull/732

.changes/unreleased/Features-20240318-033136.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: ' Add new workflow for internal patch releases'
```
emmyoop, line 2: If adapters wants these changelogs
```suggestion
body: 'Add new workflow for internal patch releases'
```

emmyoop (approved):
Approving as is for now with the understanding that more PRs will follow on after testing since none of this can be tested before the workflow gets merged.  Since this is a workflow that will be manually submitted there is little risk in this approach.

.changes/unreleased/Features-20240318-033136.yaml
aranke: I've been tagging GH actions changes with `Skip Changelog` (since these aren't shipped to end users).

Do we want to do that here too?

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+        required: false
+      package_test_command:
+        description: "Package test command"
+        type: string
+        default: "python -c \"import dbt.adapters.snowflake\""
+        required: true
+      dbms_name:
+        description: "The name of the warehouse the adapter connects to."
+        type: string
+        default: "snowflake"
```
aranke, line 22: Would this ever be a value besides `snowflake`?

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+        type: string
+        required: true
+      sha:
+        description: "The sha to use (leave empty to use latest on main)"
+        type: string
+        required: false
+      package_test_command:
+        description: "Package test command"
+        type: string
+        default: "python -c \"import dbt.adapters.snowflake\""
```
aranke, line 37: Nit; should we change this to `import dbt.adapters as adapters; assert hasattr(adapters, 'snowflake')` to make this a bit clearer?

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+    shell: bash
+
+env:
+  PYTHON_TARGET_VERSION: 3.11
+
+jobs:
+  invoke-reusable-workflow:
+    name: Build and Release Internally
+
+    uses: VersusFacit/dbt-release/.github/workflows/internal-archive-release.yml@main
```
aranke, line 56: Should we migrate this workflow to the `dbt-labs` namespace before merging?

aranke (commented):
Mostly LGTM, few nits

PR #937, opened 2024-03-19
[Fix] Update dynamic table metadata queryies to take into account new object type.

### LINKED ISSUE #934: [Regression] Snowflake is now populating a value for the relation type on dynamic tables, which is not properly reflected in the metadata queries

### Current Behavior

Querying the relation type on a dynamic table produces "BASE TABLE". This used to be empty, and was coalesced to a value unique to dynamic tables.

Querying the relation type on a dynamic table should produce "DYNAMIC TABLE".

### Steps To Reproduce

- create a dynamic table
- query the relation type, or, do anything dependent on the relation type
- note that you get behavior suggesting it's a traditional table

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core (working version):
- dbt-snowflake (working version):
- dbt-core (regression version):
- dbt-snowflake (regression version):
```

### Additional Context

We need to update the metadata query. Because it looks like the value selected matches that of a traditional table, we'll need to look for another field to differentiate the two, such as an update frequency, or the presence in a metadata table that only contains dynamic tables.

McKnight-42:
resolves #934 

### Problem
Snowflake gave dynamic tables a object type of `BASE TABLE` when it was null now we need to modify our SQL queries to build out the tables based.

### Checklist


dbt/include/snowflake/macros/catalog.sql
```diff
@@ -41,7 +41,10 @@
         table_catalog as "table_database",
         table_schema as "table_schema",
         table_name as "table_name",
-        coalesce(table_type, 'DYNAMIC TABLE') as "table_type",
+        case
+            when target_lag not null and table_type = 'BASE TABLE' THEN 'DYNAMIC TABLE'
+            else coalesce(table_type, 'BASE TABLE')
```
mikealfare, line 46: I would get rid of the `coalesce` since we should no longer need it. So just `else table_type`.

mikealfare (approved):

PR #986, opened 2024-04-25
Add entrypoint to setup.py to allow pipx installation

aranke:
### Problem

Starting in Python 3.12, we can't install packages in the global venv: https://stackoverflow.com/a/75722775

### Solution

As an alternative, we can allow users to install via `pipx` like so:

```
pipx install dbt-snowflake
```

### Manual Testing

```
❯ dbt --version
pyenv: dbt: command not found

❯ pipx install git+https://github.com/dbt-labs/dbt-snowflake.git@entrypoint_setup_py
  installed package dbt-snowflake 1.8.0b3, installed using Python 3.12.2
  These apps are now globally available
    - dbt
done! ✨ 🌟 ✨

❯ dbt --version
Core:
  - installed: 1.8.0-b3
  - latest:    1.7.13   - Ahead of latest version!

Plugins:
  - snowflake: 1.8.0b3 - Ahead of latest version!
```

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240425-143431.yaml
+++ b/.changes/unreleased/Under the Hood-20240425-143431.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add entrypoint to setup.py to allow pipx installation
+time: 2024-04-25T14:34:31.796721+02:00
+custom:
+  Author: aranke
+  Issue: ' '
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,9 @@ setup(
     url="https://github.com/dbt-labs/dbt-snowflake",
     packages=find_namespace_packages(include=["dbt", "dbt.*"]),
     include_package_data=True,
+    entry_points={
+        "console_scripts": ["dbt = dbt.cli.main:cli"],
+    },
     install_requires=[
         "dbt-common>=0.1.0a1,<2.0",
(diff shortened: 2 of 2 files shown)


mikealfare:
@aranke To maake sure I follow, in `py311` and prior, we automatically picked up the entrypoint from `dbt-core` when `pip` installed it as a dependency. But with `py312` and forward, we need to explicitly add the entrypoint in the adapter since folks will run `pip install dbt-snowflake`. Is that right?

aranke:
@mikealfare Thanks for the callout, here's a bit more detail:

Since Python 3.12 you can't install packages in the global virtual environment; so `pip install dbt-snowflake` outside a virtual environment now raises an error and exits.

There are two potential workarounds:
1. Install `dbt-snowflake` in a virtual environment (works same as before)
2. Use [`pipx`](https://pipx.pypa.io/stable/), a tool to “Install and Run Python Applications in Isolated Environments”.

Workaround 2 doesn't work right now because `pipx` looks for a console entry point in the installed package (`dbt-snowflake` here). Since this console entry point doesn't exist right now, `pipx` throws an error. This PR aims to fix this by “passing through” `dbt-core`'s console entry point to `dbt-snowflake`, thus allowing it to be installed by `pipx`.

Please let me know if I can make anything clearer still. Thanks!

PR #1022, opened 2024-05-07
Cleanup main after cutting new 1.8.latest branch

FishtownBuildBot:
This PR will fail CI until the dbt-core PR has been merged due to release version conflicts. The workflow that generated this PR also created a new branch: 1.8.latest


VersusFacit (approved):
We need to do manual cleanup on the changelog tomorrow since 1.8.0b4 exists and is not accounted ƒor because the snowflake release job went out _after_ 1.8.latest was cut.

PR #1048, opened 2024-05-16
Allow enabling/disabling stats collection on catalog table using env var

aranke:
### Problem

We're consuming a lot of memory populating the `stats` fields in `catalog`, since we have 1 record per column in `information_schema`.

### Solution

Allow disabling collection of the `stats` fields from `information_schema` to save memory.

On ad-hoc testing with my personal schema in Snowflake, this has resulted in a 20% memory decrease (690 MB → 550 MB).

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/catalog.sql
+++ b/dbt/include/snowflake/macros/catalog.sql
@@ -37,6 +37,8 @@
 
 
 {% macro snowflake__get_catalog_tables_sql(information_schema) -%}
+    {%- set enable_stats = env_var('DBT_CATALOG_TABLE_ENABLE_STATS', 'true') == 'true' -%}
+
     select
         table_catalog as "table_database",
         table_schema as "table_schema",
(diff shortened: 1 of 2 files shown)


dbt/include/snowflake/macros/catalog.sql
```diff
@@ -37,6 +37,8 @@
 
 
 {% macro snowflake__get_catalog_tables_sql(information_schema) -%}
+    {%- set enable_stats = env_var('DBT_CATALOG_TABLE_ENABLE_STATS', 'true') == 'true' -%}
```
peterallenwebb, line 40: Is there precedent for using an env_var this way in macros? I suspect the Adapters team might want us to have an official flag for this.

jtcohen6:
@aranke @peterallenwebb Thanks for the productive investigation here!

Following on this afternoon's conversation, we're going to investigate the possibility of decomposing table-level and column-level metadata/stats into two separate queries. If we decide we want that behavior to be conditional, it should be via first-class flag (CLI option / env var).

In the meantime, I'm going to close this PR.

PR #1050, opened 2024-05-17
Improve the memory efficiency of process_results() override

### LINKED ISSUE #1053: Improve Memory Efficiency of process_results() Override

### Describe the feature

The current implementation of dbt-snowflake's process_results() override holds a full additional copy of query results in memory. For queries with large result sets, such as those used by get_catalog(), this can limit dbt's ability to operate in memory-constrained environments.

peterallenwebb:
resolves #1053

### Problem

The existing override of process_results() creates a full copy of query results.

### Solution

Iterate over result rows to modify them one by one.

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/connections.py
+++ b/dbt/adapters/snowflake/connections.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from io import StringIO
 from time import sleep
-from typing import Optional, Tuple, Union, Any, List
+from typing import Any, List, Iterable, Optional, Tuple, Union
 
 import agate
 from dbt_common.clients.agate_helper import empty_table
@@ -443,25 +443,28 @@ class SnowflakeConnectionManager(SQLConnectionManager):
         split_query = snowflake.connector.util_text.split_statements(sql_buf)
         return [part[0] for part in split_query]
 
-    @classmethod
-    def process_results(cls, column_names, rows):
-        # Override for Snowflake. The datetime objects returned by
-        # snowflake-connector-python are not pickleable, so we need
-        # to replace them with sane timezones
-        fixed = []
+    @staticmethod
+    def _fix_rows(rows: Iterable[Iterable]) -> Iterable[Iterable]:
+        # See note in process_results().
         for row in rows:
             fixed_row = []
             for col in row:
                 if isinstance(col, datetime.datetime) and col.tzinfo:
                     offset = col.utcoffset()
+                    assert offset is not None
                     offset_seconds = offset.total_seconds()
-                    new_timezone = pytz.FixedOffset(offset_seconds // 60)
+                    new_timezone = pytz.FixedOffset(int(offset_seconds // 60))
                     col = col.astimezone(tz=new_timezone)
                 fixed_row.append(col)
 
-            fixed.append(fixed_row)
+            yield fixed_row
 
-        return super().process_results(column_names, fixed)
+    @classmethod
+    def process_results(cls, column_names, rows):
+        # Override for Snowflake. The datetime objects returned by
+        # snowflake-connector-python are not pickleable, so we need
+        # to replace them with sane timezones.
+        return super().process_results(column_names, cls._fix_rows(rows))
 
     def execute(
         self, sql: str, auto_begin: bool = False, fetch: bool = False, limit: Optional[int] = None
--- a/.changes/unreleased/Under the Hood-20240517-143743.yaml
+++ b/.changes/unreleased/Under the Hood-20240517-143743.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Improve memory efficiency of the process_results() override.
+time: 2024-05-17T14:37:43.7414-04:00
+custom:
+  Author: peterallenwebb
+  Issue: "1053"


dbt/adapters/snowflake/connections.py
```diff
@@ -443,25 +443,28 @@ def _split_queries(cls, sql):
...
+                    new_timezone = pytz.FixedOffset(int(offset_seconds // 60))
                     col = col.astimezone(tz=new_timezone)
                 fixed_row.append(col)
 
-            fixed.append(fixed_row)
+            yield fixed_row
 
-        return super().process_results(column_names, fixed)
+    @classmethod
+    def process_results(cls, column_names, rows):
```
aranke, line 463: Do we even need this method anymore, or can we delete it?

dbt/adapters/snowflake/connections.py
```diff
@@ -443,25 +443,28 @@ def _split_queries(cls, sql):
...
+    @staticmethod
+    def _fix_rows(rows: Iterable[Iterable]) -> Iterable[Iterable]:
+        # See note in process_results().
         for row in rows:
             fixed_row = []
             for col in row:
                 if isinstance(col, datetime.datetime) and col.tzinfo:
                     offset = col.utcoffset()
+                    assert offset is not None
                     offset_seconds = offset.total_seconds()
```
aranke, line 455: Nit: `offset_seconds = offset.total_seconds() or 0`

dbt/adapters/snowflake/connections.py
```diff
@@ -443,25 +443,28 @@ def _split_queries(cls, sql):
...
-                    new_timezone = pytz.FixedOffset(offset_seconds // 60)
+                    new_timezone = pytz.FixedOffset(int(offset_seconds // 60))
```
aranke, line 456: `pytz` is no longer required as of 3.9+, let's try to remove the dependency here?

https://github.com/stub42/pytz/blob/master/src/README.rst#introduction

dbt/adapters/snowflake/connections.py
```diff
@@ -443,25 +443,28 @@ def _split_queries(cls, sql):
...
-        fixed = []
+    @staticmethod
+    def _fix_rows(rows: Iterable[Iterable]) -> Iterable[Iterable]:
+        # See note in process_results().
         for row in rows:
             fixed_row = []
             for col in row:
                 if isinstance(col, datetime.datetime) and col.tzinfo:
                     offset = col.utcoffset()
+                    assert offset is not None
```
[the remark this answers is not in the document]
peterallenwebb: Correct. The check for col.tzinfo is what prevents this from ever crashing, since it means offset will never be None, but mypy doesn't realize that, and this keeps it happy.

dbt/adapters/snowflake/connections.py
peterallenwebb: I did this in Python to check some things for myself:

```
>>> from typing import Iterator

>>> def my_generator():
...     for i in [1, 2, 3, 4, 5]:
...         yield i
...

>>> def get_my_generator():
...     return my_generator()
...

>>> type(my_generator())
<class 'generator'>

>>> type(get_my_generator())
<class 'generator'>

>>> isinstance(my_generator(), Iterable)
True

>>> isinstance(my_generator(), Iterator)
True
```

Conclusions:

The `return` at the end of process_results() is correct, since it is returning a generator (which is an instance of Iterable), not an element of the generator.

dbt/adapters/snowflake/connections.py
```diff
@@ -443,25 +443,28 @@ def _split_queries(cls, sql):
...
         return [part[0] for part in split_query]
 
-    @classmethod
-    def process_results(cls, column_names, rows):
-        # Override for Snowflake. The datetime objects returned by
-        # snowflake-connector-python are not pickleable, so we need
-        # to replace them with sane timezones
-        fixed = []
+    @staticmethod
+    def _fix_rows(rows: Iterable[Iterable]) -> Iterable[Iterable]:
```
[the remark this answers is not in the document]
peterallenwebb: @mikealfare This approach makes sense to me, but I would also like to get this change out tomorrow. Would you be open to dealing with the new override structure in a follow-up?

Regarding the return type, every Iterator is also Iterable, so I believe it is correct. And it is also as non-specific as possible, which I would argue is good, since it gives us the widest degree of flexibility in what we accept and return.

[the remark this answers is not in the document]
peterallenwebb: This is what I mean:

```
>>> issubclass(Iterator, Iterable)
True
```

dbt/adapters/snowflake/connections.py
mikealfare: Interesting, that's not how I would have thought that worked, but it does. TIL

dbt/adapters/snowflake/connections.py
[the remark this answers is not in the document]
mikealfare: > Would you be open to dealing with the new override structure in a follow-up?

Yes, this is fine. We should probably discuss the generic approach before doing it anyway. It's just a low hanging beam that we constantly bonk our heads on, so I always tend towards fixing it.

> Regarding the return type, every Iterator is also Iterable, so I believe it is correct. And it is also as non-specific as possible, which I would argue is good, since it gives us the widest degree of flexibility in what we accept and return.

Sure, both are correct. I didn't intend to say that `Iterable` is incorrect. I was simply arguing that `Iterator` better reflects what's being done in this method and the story in general. I agree that argument types should be more general to allow for more flexible implementations, as long as they capture the intended requirements of the argument. Along those same lines, the return type should capture the intended usage of the result. I understood that intended usage to be something that was lazy loaded to improve memory usage, especially given the usage of `yield` and the fact that this method will never be overridden by a downstream application. If that's not the case, then `Iterable` is fine.

mikealfare (approved):

PR #1058, opened 2024-05-21
SPIKE: external table as model materialization

dataders:
related: https://github.com/dbt-labs/dbt-adapters/pull/220


amychen1776:
Closing this for now because this is not planned work and we have revisited our thinking about external tables

PR #1060, opened 2024-05-22
[Bug] Use fully qualified names in rename for tables and views

### LINKED ISSUE #1031: [Bug] Use qualified names in `snowflake__get_rename_x_sql` macros

### Current Behavior

The `snowflake__get_rename_x_sql` macros are using unqualified names for the rename target. This ultimately led to an error in #1016 — although I'm glad it did, because if it hadn't, it might have masked the real underlying issue, which is that dbt can no longer recognizes dynamic tables based on the output of `show terse objects` and is full-refreshing them on every run.

Still, I would expect that these macros return:

```sql
alter table "ANALYTICS"."DBT_JCOHEN_DEV"."MY_DYNAMIC_TABLE" rename to "ANALYTICS"."DBT_JCOHEN_DEV".MY_DYNAMIC_TABLE__dbt_backup;
```
Instead:
```sql
alter table "ANALYTICS"."DBT_JCOHEN_DEV"."MY_DYNAMIC_TABLE" rename to MY_DYNAMIC_TABLE__dbt_backup;
```

### Expected Behavior

```sql
{%- macro snowflake__get_rename_table_sql(relation, new_name) -%}
    alter table {{ relation }} rename to {{ relation.incorporate(path={"identifier": new_name}) }}
{%- endmacro -%}

{%- macro snowflake__get_rename_view_sql(relation, new_name) -%}
    alter view {{ relation }} rename to {{ relation.incorporate(path={"identifier": new_name}) }}
{%- endmacro -%}
```

### Steps To Reproduce

Reproduction case in #1016 — the specific error actually surfaces because the temp relation being dropped (qualified name) does not match the temp relation being created (unqualified name)

### Relevant log output

```shell
08:31:27  On model.my_dbt_project.foo_2024_03_enabled: /* {"app": "dbt", "dbt_version": "1.7.14", "profile_name": "all", "target_name": "sf", "node_id": "model.my_dbt_project.foo_2024_03_enabled"} */
-- get the standard backup name
    

    -- drop any pre-existing backup
    

    
        drop table if exists "DEVELOPMENT_JYEO"."DBT_JYEO"."FOO_2024_03_ENABLED__dbt_backup" cascade

    

;
08:31:27  Opening a new connection, currently in state closed
08:31:29  SQL status: SUCCESS 1 in 2.0 seconds
08:31:29  Using snowflake connection "model.my_dbt_project.foo_2024_03_enabled"
08:31:29  On model.my_dbt_project.foo_2024_03_enabled: /* {"app": "dbt", "dbt_version": "1.7.14", "profile_name": "all", "target_name": "sf", "node_id": "model.my_dbt_project.foo_2024_03_enabled"} */
alter table "DEVELOPMENT_JYEO"."DBT_JYEO"."FOO_2024_03_ENABLED" rename to FOO_2024_03_ENABLED__dbt_backup;
08:31:29  Snowflake adapter: Snowflake query id: 01b413bf-0804-98fe-000d-3783323f397a
08:31:29  Snowflake adapter: Snowflake error: 002002 (42710): SQL compilation error:
Object 'FOO_2024_03_ENABLED__DBT_BACKUP' already exists.
```

### Environment

```markdown
- OS: 14.4.1
- Python: 3.10.11
- dbt-core: 1.8.0
- dbt-snowflake: 1.8.0
```

### Additional Context

_No response_

mikealfare:
resolves #1031

### Problem

The rename macro for tables and views uses only the identifier. It's expected to use the fully qualified name.

### Solution

Update these macros to incorporate the new name into the source relation, thereby preserving the database and schema.

### Checklist


dbt/include/snowflake/macros/relations/table/rename.sql
```diff
@@ -1,3 +1,3 @@
 {%- macro snowflake__get_rename_table_sql(relation, new_name) -%}
-    alter table {{ relation }} rename to {{ new_name }}
+    alter table {{ relation }} rename to {{ relation.incorporate(path={"identifier": new_name}) }}
```
jtcohen6, line 2: As a user pointed out in [this comment](https://github.com/dbt-labs/dbt-snowflake/commit/cea7f0589c340fee3a67320cdb7e148454720148#r142183494), it isn't clear right now whether the type signature of `get_rename_x_sql` is:
```
get_rename_table_sql(relation: BaseRelation, new_name: str)
get_rename_table_sql(relation: BaseRelation, new_name: BaseRelation)
```
This change will break if someone is currently passing a `Relation` object into the second argument. Do you think we should add conditional handling based on the type of `new_name`?

mikealfare: My short answer is that I'd rather not overload the argument.

While I specifically chose the suffix `name` to distinguish from `relation`, that's an implicit assumption. We could adopt the practice of adding the equivalent of a docstring to macros that we expect users to use (or maybe just all of them for our own sake). I like this idea better than what we're currently doing. If we instead add a conditional that handles both a relation and a string, then we're advertising that both are acceptable when we really don't expect a relation.

I have a stronger concern from the user's second comment, that this takes away the ability to use this macro to move an object to a different schema by supplying a fully qualified name. The Snowflake docs on [ALTER TABLE](https://docs.snowflake.com/en/sql-reference/sql/alter-table) suggest that the default for <new_table_name> is just the identifier, which aligns with my experience in other databases. They go on further in the parameters section to mention that you could use an ALTER/RENAME statement to move an object from one database/schema to another. That's possible with the current implementation, but not so with this version.

So talking that out, I actually think we should not make this change at all. However, it's certainly worth adding the docstring to make it clearer to users what this macro expects for arguments and how it's expected to be used. I would also keep the tests that were added and update them to whatever we choose to go with. I'm definitely interested in your thoughts on this before moving forward as this is ultimately your call.

jtcohen6: @mikealfare Thanks for thinking this through!
- I agree we shouldn't overload the type of this macro, it should be either `str` or `Relation` (and `str` to avoid a breaking change with current expected type)
- If I had my druthers, we would _always_ use fully-qualified relation names, to avoid this and related issues. (The alternative is to [run `use` statements before every single query](https://github.com/dbt-labs/dbt-snowflake/issues/1051))
- Are we ever "renaming" an object to change more than just its identifier (i.e. also changing its database/schema)? I don't think this is something we ever(?) do within dbt today. In theory, we could do this in the future, for dynamic tables and incremental models, as a form of database migration, if we know the object exists in its old location and only the configured database/schema has changed — a form of "drift detection" — but the current behavior is to presume that the object _doesn't_ already exist. If we're only ever renaming object identifiers, then reusing the database/schema from the old relation name seems reasonable.

I might have been confused about the specific bug that surfaced with dynamic tables, which motivated me to open https://github.com/dbt-labs/dbt-snowflake/issues/1031. I agree the actual underlying issue was the `2024_03` bundle change to `show terse objects`, which led dbt to think a DT was actually a table, and try to drop + recreate it accordingly. But it was during that drop + recreate (= create-alter-rename-swap-drop), the backup DT was supposed to be getting dropped, but dbt wasn't dropping it correctly, because it wasn't actually renaming them correctly.

Basically, something like this:
```sql
-- implicit each time we open a new connection, with the default schema set to {{ target.schema }}
use schema analytics.dbt_jcohen_dev;

-- imagine a table configured with a custom schema
create schema if not exists analytics.dbt_jcohen_other;
create or replace table analytics.dbt_jcohen_other.my_table as (select 1 as id);

-- a new connection
use schema analytics.dbt_jcohen_dev;
alter table analytics.dbt_jcohen_other.my_table rename to my_table__dbt_backup;
drop table analytics.dbt_jcohen_other.my_table__dbt_backup;
```
This fails because the table was renamed to `analytics.dbt_jcohen_dev.my_table__dbt_backup`, NOT `analytics.dbt_jcohen_other.my_table__dbt_backup`. In the meantime, the renamed DT keeps "hanging out" and is never actually dropped.

jtcohen6: _notes from live conversation with @mikealfare_

Two options here:
- Current proposed change. We always rename to a fully qualified identifier name. The intended scope of `get_rename_sql` is _only_ to update the `identifier` of an object (view/table/etc). It does not support moving objects into different databases/schemas. If we need to do the latter in the future, we'll implement some other macro (`get_rename_relation_sql`).
- Or: `get_rename_sql` accepts `Union[str, Relation]` as the type for `new_name`, and handles either with conditional logic. This feels like too low-level a place to be doing the type checking, and we'll be doing it several times in different places.

---

Proposal: The [top-level `get_rename_sql` macro](https://github.com/dbt-labs/dbt-adapters/blob/a2292c8ec76e4c9f03869ad95817a2ad82dfb34b/dbt/include/global_project/macros/relations/rename.sql#L3) should accept two arguments: `relation` (which is a `Relation`) and `new_name`, which is either:
- A string, representing just an updated `identifier` for `relation` (first argument). We expect this is >95% of the expected usage of this macro, and so it should be easy to use it in this way.
- A `Relation` object, which it will convert to a string (via `render()`) representing that relation's fully qualified name.
- It is not supported to pass a `string` like `"different_database.different_schema.different_identifier"`. If a user wants to leverage `get_rename_sql` for _moving_ objects to different database/schema, they need to build a `Relation` object and pass that in.

```
{%- macro get_rename_sql(relation, new_name) -%}
    {{- log('Applying RENAME to: ' ~ relation) -}}
    {% set new_name = relation.incorporate(path={"identifier": new_name}).render() if new_name is string else new_name.render() %}
    {{- adapter.dispatch('get_rename_sql', 'dbt')(relation, new_name) -}}
{%- endmacro -%}
```

In [this example](https://github.com/dbt-labs/dbt-adapters/blob/a2292c8ec76e4c9f03869ad95817a2ad82dfb34b/dbt/include/global_project/macros/relations/create_backup.sql#L15), whether we make the second argument `backup_relation.identifier` ("happy path") OR just `backup_relation`, we get the same result. In that way, this is a backward-compatible change and yields what the user would expect.

What feels less good? The lower-level rename macros to which it dispatches should accept `relation` (which is a `Relation`) and `new_name` (which is always a string, representing a fully qualified relation name).

We're going to need to make some choice here that feels less-than-ideal (or has some trade-off), so we're going to spend a little (not too much) time comparing two potential paths forward and then pick one.

mikealfare: @jtcohen6 Please see the new files that I pushed to this PR:
- `_rename_approach_0.sql` - current state
- `_rename_approach_1.sql` - this PR's approach
- `_rename_approach_2.sql` - the approach from our discussion
- `_rename_approach_3.sql` - a compromise of 1 and 2
- `_rename_usage.sql` - two impacted use cases of these changes

I combined macros from different files and packages in one spot so that we don't need to flip back and forth to follow the flow. I also highlighted the change with a comment. Most of the code really isn't changing, but it helps to see where the changed code is. My new proposal is option 3.

I have two more outcomes from this exercise:
- it's probably worth documenting this flow (e.g. `_rename_approach_0.sql`) somewhere for future discussions
- I really like the idea of adding python-like docstrings to macros to communicate intent and expected inputs; this would be useful for adapter maintainers who need to implement these macros, and end users who may use these macros in their custom project macros

jtcohen6: @mikealfare Thanks for the thorough walkthrough! So the idea behind approach 4 is:
- Snowflake's behavior here is unique, in its treatment of renaming objects based on the defaults specified in the connection
- Therefore, we should make this change _within_ `dbt-snowflake`, to pass the entire rendered relation name (fully qualified) from the backup/intermediate rename macros

I think I buy your reasoning, that this approach is:
- the least disruptive to existing functionality, other adapters, and the base spec
- justifiable as an isolated change to `dbt-snowflake` because of this unique(?) behavior on Snowflake

colin-rogers-dbt (approved):

Issue #1059, opened 2024-05-22
[Feature] add functional test to cover new cover new custom config for tests.


McKnight-42:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-snowflake functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

After merging prs made off #https://github.com/dbt-labs/dbt-core/issues/10197 and #https://github.com/dbt-labs/dbt-core/issues/10198 we need to add new tests to dbt-snowflake to handle the custom config options now available 

suggested config `snowflake_warehouse`

### Describe alternatives you've considered

suggestion from previous pr is to 

https://github.com/dbt-labs/dbt-core/issues/10197 is complete: Add to [the functional tests in dbt-snowflake here](https://github.com/dbt-labs/dbt-snowflake/blob/5da6fdbfd0895cf86c49dd2b95d89d3d2a57aa70/tests/functional/warehouse_test/test_warehouses.py#L8) here. A test configured with a snowflake_warehouse that does not exist should fail. Today, it doesn't, because it doesn't actually use this config.

### Are you interested in contributing this feature?

yes

PR #1070, opened 2024-06-04
[Feature] update pre/post_model_hooks for tests

McKnight-42:
resolves #https://github.com/dbt-labs/dbt-snowflake/issues/1059

### Problem

We want to be able to correctly handle arbitrary configs in tests correctly, for models this is currently done with `pre/post_model_hook` and for snowflake this is mainly for swapping `snowflake_wareshouse` values, currently we don't actually recognize this config in during tests. by adding updating `dbt-core` methods we have opened up the ability to pass these into tests correctly but must make specific modifications in the adapters themselves.

asks:
1. update for logging
2. add tests

### Solution
we can use AdapterLogger to add loggger.info as we need in the `model_hook` methods

add a new tests to handle scenarios i.e we update `snowflake_warehouse` to a invalid warehouse.

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -7,6 +7,7 @@ from dbt.adapters.base.impl import AdapterConfig, ConstraintSupport
 from dbt.adapters.base.meta import available
 from dbt.adapters.capability import CapabilityDict, CapabilitySupport, Support, Capability
 from dbt.adapters.sql import SQLAdapter
+from dbt.adapters.events.logging import AdapterLogger
 from dbt.adapters.sql.impl import (
     LIST_SCHEMAS_MACRO_NAME,
     LIST_RELATIONS_MACRO_NAME,
@@ -18,6 +19,8 @@ from dbt.adapters.snowflake import SnowflakeColumn
 from dbt_common.contracts.constraints import ConstraintType
 from dbt_common.exceptions import CompilationError, DbtDatabaseError, DbtRuntimeError
 from dbt_common.utils import filter_null_values
+
+logger = AdapterLogger("Snowflake")
 
 
 @dataclass
@@ -99,15 +102,20 @@ class SnowflakeAdapter(SQLAdapter):
     def pre_model_hook(self, config: Mapping[str, Any]) -> Optional[str]:
         default_warehouse = self.config.credentials.warehouse
         warehouse = config.get("snowflake_warehouse", default_warehouse)
+        logger.info(f"Running pre_model_hook with config: {config}")
+        logger.info(f"Default warehouse: {default_warehouse}, Selected warehouse: {warehouse}")
         if warehouse == default_warehouse or warehouse is None:
             return None
         previous = self._get_warehouse()
         self._use_warehouse(warehouse)
+        logger.info(f"Changed warehouse from {previous} to {warehouse}")
         return previous
 
     def post_model_hook(self, config: Mapping[str, Any], context: Optional[str]) -> None:
+        logger.info(f"Running post_model_hook with config: {config} and context: {context}")
         if context is not None:
             self._use_warehouse(context)
+            logger.info(f"Restored warehouse to {context}")
 
     def list_schemas(self, database: str) -> List[str]:
         try:


McKnight-42:
running test without the `expect_pass` shows this in logs 

```
use warehouse DBT_TEST_DOES_NOT_EXIST
[0m11:01:25.737689 [debug] [Thread-22 ]: Snowflake adapter: Snowflake query id: 01b4cf21-0804-bc1d-000d-3783363da962
[0m11:01:25.739334 [debug] [Thread-22 ]: Snowflake adapter: Snowflake error: 002043 (02000): SQL compilation error:
Object does not exist, or operation cannot be performed.
```

dbt/adapters/snowflake/impl.py
```diff
@@ -99,15 +102,20 @@ def _use_warehouse(self, warehouse: str):
     def pre_model_hook(self, config: Mapping[str, Any]) -> Optional[str]:
         default_warehouse = self.config.credentials.warehouse
         warehouse = config.get("snowflake_warehouse", default_warehouse)
+        logger.info("Running pre_model_hook:")
+        logger.info(f"Default warehouse: {default_warehouse}, Selected warehouse: {warehouse}")
         if warehouse == default_warehouse or warehouse is None:
             return None
         previous = self._get_warehouse()
         self._use_warehouse(warehouse)
+        logger.info(f"Changed wareho use from {previous} to {warehouse}")
```
McKnight-42, line 111: Do we want to put these logs behind condition of warehouses being different?

jtcohen6: @McKnight-42 I don't think we should add any new logging here. The actual queries being run (`select current_warehouse() as warehouse` + `use warehouse different_warehouse`) are already appearing in the logs.

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -90,3 +89,25 @@ def test_snowflake_override_ok(self, project):
...
+            "invalid_warehouse.sql": project_config_models__invalid_warehouse_sql,
+        }
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "config-version": 2,
+            "models": {
+                "test": {
+                    "snowflake_warehouse": "DBT_TEST_DOES_NOT_EXIST",
```
jtcohen6, line 107: Yes, nicely done - we should expect the dbt test to try using a Snowflake warehouse that doesn't exist (`use warehouse DBT_TEST_DOES_NOT_EXIST`), where previously it wouldn't, and so it should fail with `Object does not exist, or operation cannot be performed.`

McKnight-42: went ahead and also added a assert to catch if that error message stops showing up.

mikealfare: For the sake of completeness, can we add a positive test case that uses this config option, just to make sure it's setting it? The error message we're getting in the negative test case is pretty general.

dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
-# install latest changes in dbt-core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
+# install latest changes in dbt-core"
+git+https://github.com/dbt-labs/dbt-core.git@mcknight/10198#egg=dbt-core&subdirectory=core
```
mikealfare, line 2: Reminder to revert this prior to merge.

tests/functional/warehouse_test/test_warehouses.py
McKnight-42: added a new test

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -3,35 +3,39 @@
...
 {{ config(materialized='table') }}
 select '{{ env_var("SNOWFLAKE_TEST_ALT_WAREHOUSE", "DBT_TEST_ALT") }}' as warehouse
+"""
 
+project_config_models__invalid_warehouse_sql = """
+{{ config(materialized='table') }}
+select current_warehouse() as warehouse
+"""
+
+project_config_models__valid_warehouse_sql = """
```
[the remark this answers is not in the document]
McKnight-42: good point, think i'm just in a flow from previous pr of splitting concerns will rename original and remove the second as it can be used for both

tests/functional/warehouse_test/test_warehouses.py
```diff
@@ -90,3 +89,47 @@ def test_snowflake_override_ok(self, project):
...
+            },
+        }
+
+    def test_snowflake_override_invalid(self, project):
+        result = run_dbt(["run", "--models", "invalid_warehouse"], expect_pass=False)
+        assert "Object does not exist, or operation cannot be performed" in result[0].message
+
+
+class TestValidConfigWarehouse:
+    @pytest.fixture(scope="class")
```
ChenyuLInx, line 116: I think you just need to define this and the other model in outside of all classes once.

ChenyuLInx (approved):
One nit otherwise LGTM

ChenyuLInx:
This is just a test I don't think it is a feature.

McKnight-42:
> This is just a test I don't think it is a feature.

currently labeled as such because it ties with the core prs https://github.com/dbt-labs/dbt-core/pull/10258 and https://github.com/dbt-labs/dbt-core/pull/10245 but yeah could possibly just be a under the hood update.

colin-rogers-dbt (approved):

PR #1072, opened 2024-06-05
[Bug] Silent SSO token expiration

### LINKED ISSUE #851: [Bug] Snowflake SSO Token expiration not surfaced in logs

### Current Behavior

SSO auth is silently failing after token expiration. 

### Expected Behavior

An auth failure exception is raised

### Steps To Reproduce

On Version 1.5.9 ->
Token expired for Snowflake SSO  ->
Run ‘dbt run’ command in the IDE

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-snowflake:
```

### Additional Context

_No response_

mikealfare:
resolves #851

### Problem

Expired SSO tokens fail silently.

### Solution

Catch the specific exception for an expired SSO token and raise it so the user is aware of why the connection is failing.

### Checklist

### DIFF
--- a/tests/functional/oauth/test_oauth.py
+++ b/tests/functional/oauth/test_oauth.py
@@ -31,9 +31,11 @@ Update CI providers and test.env with the new values (If you kept the security
 integration the same, just the refresh token changed)
 """
 
-import pytest
 import os
-from dbt.tests.util import run_dbt, check_relations_equal
+
+from dbt.adapters.exceptions.connection import FailedToConnectError
+from dbt.tests.util import check_relations_equal, run_dbt
+import pytest
 
 
 _MODELS__MODEL_1_SQL = """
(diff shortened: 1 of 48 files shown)


McKnight-42:
looking into the further up swap to a `Database Error`

looking at `open` method in connections.py we get to 

```
retryable_exceptions = [
            InternalError,
            InternalServerError,
            ServiceUnavailableError,
            GatewayTimeoutError,
            RequestTimeoutError,
            BadGatewayError,
            OtherHTTPRetryableError,
            BindUploadError,
        ]
        # these two options are for backwards compatibility
        if creds.retry_all:
            retryable_exceptions = [Error]
        elif creds.retry_on_database_errors:
            retryable_exceptions.insert(0, DatabaseError)

        return cls.retry_connection(
            connection,
            connect=connect,
            logger=logger,
            retry_limit=creds.connect_retries,
            retry_timeout=timeout if timeout is not None else exponential_backoff,
            retryable_exceptions=retryable_exceptions,
        )
```

falling back to dbt-adapters and looking at the method defined there as part of `BaseConnectionManager` failure seems to happen here

```
except Exception as e:
            connection.handle = None
            connection.state = ConnectionState.FAIL  # type: ignore
            raise FailedToConnectError(str(e))
```
applying breakpoint here and checking type and str of `e` shows:
```
type(e)
<class 'dbt.adapters.exceptions.connection.FailedToConnectError'>
str: str(e)
'Database Error\n  This error occurs when authentication has expired. Please reauth with your auth provider.'
```


looking into where the DatabaseError is joined on

tests/functional/oauth/test_oauth.py
```diff
@@ -88,3 +90,28 @@ def models(self):
     def test_snowflake_basic(self, project):
         run_dbt()
         check_relations_equal(project.adapter, ["MODEL_3", "MODEL_4"])
+
+
+class TestSnowflakeOAuthExpiration:
```
colin-rogers-dbt, line 95: Looking at the test run logs I think the issue here is that as part of test setup we are creating a connection with the database so it's failing in test setup and not in the `test_token_expiration` method

McKnight-42: updated the test, it hits the error then for some reason continues on into a RunTimeError as well

colin-rogers-dbt (approved):

PR #1085, opened 2024-06-14
[Tech Debt] Fix dynamic table tests

### LINKED ISSUE #1084: [Tech Debt] Fix DT testing that should be automated

### Current Behavior

Many of the current DT tests need to be run manually.

### Expected Behavior

All tests should run automatically.

### Steps To Reproduce

You can see that there are parts of tests that are commented out that should run.

mikealfare:
resolves #1084

### Problem

We were manually running some integration tests.

### Solution

Update the tests so that they could run automatically.

### Checklist

### DIFF
--- a/tests/functional/adapter/dynamic_table_tests/test_dynamic_tables_changes.py
+++ b/tests/functional/adapter/dynamic_table_tests/test_dynamic_tables_changes.py
@@ -25,21 +25,15 @@ from tests.functional.adapter.dynamic_table_tests.utils import (
 
 class SnowflakeDynamicTableChanges:
     @staticmethod
-    def check_start_state(adapter, dynamic_table):
-        """
-        This needs to be done manually for now until we fix the test suite's runner. The test suite's
-        runner cannot run queries with multiple statements. Snowflake's metadata is all behind `show`
-        and `describe` calls that require a second call to fetch the results; hence, the results
-        cannot be fetched.
-        """
-        assert query_target_lag(adapter, dynamic_table) is None == "120 seconds"
-        assert query_warehouse(adapter, dynamic_table) is None == "DBT_TESTING"
+    def check_start_state(project, dynamic_table):
+        assert query_target_lag(project, dynamic_table) is None == "2 minutes"
+        assert query_warehouse(project, dynamic_table) is None == "DBT_TESTING"
 
     @staticmethod
     def change_config_via_alter(project, dynamic_table):
         initial_model = get_model_file(project, dynamic_table)
         new_model = initial_model.replace(
-            "target_lag='120        seconds'", "target_lag='5   minutes'"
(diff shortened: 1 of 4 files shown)


tests/functional/adapter/dynamic_table_tests/test_dynamic_tables_changes.py
```diff
@@ -103,6 +95,9 @@ def setup(self, project, my_dynamic_table):
         # the tests touch these files, store their contents in memory
         initial_model = get_model_file(project, my_dynamic_table)
 
+        # verify the initial settings are correct in Snowflake
+        self.check_start_state(project, my_dynamic_table)
```
mikealfare, line 99: This used to be commented at the beginning of every test. It was moved here and will run automatically. So it looks like the comment was removed from each test, but it as effectively just uncommented.

tests/functional/adapter/dynamic_table_tests/test_dynamic_tables_changes.py
```diff
@@ -131,17 +133,16 @@ class TestSnowflakeDynamicTableChangesApply(SnowflakeDynamicTableChanges):
     def project_config_update(self):
         return {"models": {"on_configuration_change": OnConfigurationChangeOption.Apply.value}}
 
-    def test_change_is_applied_via_alter(self, project, adapter, my_dynamic_table):
-        """
-        See above about the two commented assertions. In the meantime, these have been validated manually.
-        """
-        # self.check_start_state(adapter, my_dynamic_table)
```
mikealfare, line 138: It looks like this was deleted, but it was just uncommented. See my GH comment on line 99. This is true of all lines like this, but I'm leaving one comment for the sake of brevity.

tests/functional/adapter/dynamic_table_tests/test_dynamic_tables_changes.py
```diff
@@ -131,17 +133,16 @@ class TestSnowflakeDynamicTableChangesApply(SnowflakeDynamicTableChanges):
...
+    def test_change_is_applied_via_alter(self, project, my_dynamic_table):
 
+        # update the settings
         self.change_config_via_alter(project, my_dynamic_table)
         _, logs = run_dbt_and_capture(["--debug", "run", "--models", my_dynamic_table.name])
 
-        # self.check_state_alter_change_is_applied(adapter, my_dynamic_table)
+        # verify the updated settings are correct in Snowflake
+        self.check_state_alter_change_is_applied(project, my_dynamic_table)
```
mikealfare, line 143: We're now checking in the database to verify the settings were applied. Even though that is a stronger check than checking the logs, we still want to check the logs so that we know how the change was made. If a change that was supposed to be applied via an alter is actually applied via a drop/create, that's not passing.

colin-rogers-dbt (approved):

PR #1100, opened 2024-06-28
[Bug] Fix scenarios where `--empty` flag pushes `limit 0` into metadata queries

### LINKED ISSUE #1033: [Bug] --empty doesn't work when using get_columns_in_relation

### Current Behavior

The `--empty` flag changes the relation name to be something like this:
`(select * from DATABASE_NAME.SCHEMA_NAME.TABLE_NAME where false limit 0)`
This becomes an issue when the adapter `get_columns_in_relation` function is called as it tries to do something like this, which is invalid snowflake sql:
`describe table (select * from DATABASE_NAME.SCHEMA_NAME.TABLE_NAME where false limit 0)`

This means we get the following sql errors when using functions like `dbt_utils.star`:
`Syntax error: unexpected '('. (line 1)`

### Expected Behavior

anywhere where the `{{ relation }}` is used after a describe statement, this should be handled differently for the --empty flag to not be the subquery, but instead use the default original relation.
Like here:
https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/include/snowflake/macros/adapters.sql#L30

### Steps To Reproduce

1. dbt-snowflake 1.8
2. running with the `--empty` command
3. on any model which includes the `{{ dbt_utils.star() }}` macro

### Relevant log output

_No response_

### Environment

```markdown
- OS: Linux amd64
- Python: 3.10
- dbt-core: 1.8.0
- dbt-snowflake: 1.8.0
```

### Additional Context

_No response_

mikealfare:
resolves #1033

### Problem

We use `{{ relation }}` to render the relation name in metadata queries. However, we implemented the `--empty` flag to add `limit 0` when `{{ relation }}` is used. We wind up with queries like:
```sql
describe table (select * from my_relation limit 0)
```
which is not valid sql.

### Solution

- Use `{{ relation.render() }}` in these scenarios (there are many)
- Add a test case that confirms we can use these macros while using the `--empty` flag

### Implementation

1. Merge this PR demonstrating that these tests work (we`re here)
2. Move the bulk of this PR into `dbt-tests-adapter`
3. Create a final PR that removes most of this and uses the `dbt-tests-adapter` tests

### Checklist

### DIFF
--- a/tests/functional/adapter/empty/test_empty.py
+++ b/tests/functional/adapter/empty/test_empty.py
@@ -21,9 +21,14 @@ class TestMetadataWithEmptyFlag:
     @pytest.fixture(scope="class")
     def models(self):
         return {
+            "schema.yml": _models.SCHEMA,
             "control.sql": _models.CONTROL,
             "get_columns_in_relation.sql": _models.GET_COLUMNS_IN_RELATION,
             "alter_column_type.sql": _models.ALTER_COLUMN_TYPE,
+            "alter_relation_comment.sql": _models.ALTER_RELATION_COMMENT,
+            "alter_column_comment.sql": _models.ALTER_COLUMN_COMMENT,
+            "alter_relation_add_remove_columns.sql": _models.ALTER_RELATION_ADD_REMOVE_COLUMNS,
+            "truncate_relation.sql": _models.TRUNCATE_RELATION,
         }
 
     @pytest.fixture(scope="class", autouse=True)
@@ -36,6 +41,10 @@ class TestMetadataWithEmptyFlag:
             "control",
             "get_columns_in_relation",
             "alter_column_type",
+            "alter_relation_comment",
+            "alter_column_comment",
+            "alter_relation_add_remove_columns",
+            "truncate_relation",
         ],
     )
     def test_run(self, project, model):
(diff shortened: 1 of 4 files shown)


tests/functional/adapter/empty/test_empty.py
```diff
@@ -7,3 +11,41 @@ class TestSnowflakeEmpty(BaseTestEmpty):
...
+            "control",
+            "get_columns_in_relation",
+            "alter_column_type",
+            "alter_relation_comment",
+            "alter_column_comment",
+            "alter_relation_add_remove_columns",
+            "truncate_relation",
+        ],
+    )
+    def test_run(self, project, model):
```
mikealfare, line 50: Some of these tests pass even without fixing `{{ relation }}` to be `{{ relation.render() }}`. I think it has something to do with whether the template returns just the sql or whether it does the call. I added the tests in case this changes in the future, and updated the macros to use `{{ relation.render() }}`.

tests/functional/adapter/empty/test_empty.py
mikealfare, line 1: Most of this will move into `dbt-tests-adapter` and then imported here afterwards.

McKnight-42 (commented):
This is looking good to me

colin-rogers-dbt (approved):

pempey:
@mikealfare When might one expect these change in a release?  Or is there a way to test these changes on our system before a release?

pempey:
I will ask my question to a larger audience: @colin-rogers-dbt @McKnight-42 @jtcohen6  When might one expect these change in a release? Or is there a way to test these changes on our system before a release?

PR #1107, opened 2024-07-05
Fix #1102: Use `show ... starts with` instead of `show ... like` in `_show_object_metadata`

### LINKED ISSUE #1102: [Bug] Use `show ... starts with` instead of `show ... like` in `_show_object_metadata`

### Current Behavior

`show objects` queries can get slow and take up to 3s

### Expected Behavior

`show objects` queries take optimal 300ms

### Steps To Reproduce

1. Run many individual `show objects` queries in sequence

### Relevant log output

_No response_

### Environment

_No response_

### Additional Context

Solution from Snowflake: https://dbt-labs.slack.com/archives/C04R0T4H8ES/p1719874443806149?thread_ts=1719586466.620999&cid=C04R0T4H8ES

aranke:
Resolves #1102 

### Problem

We want to use `show...starts with` instead of `show...like` in `_show_object_metadata` for faster query times, as per Snowflake guidance.

### Solution

1. Modify the `snowflake__show_object_metadata` macro
2. Since `starts with` is case-sensitive, convert cases within the relation before sending off to macro; this is done via the new `as_case_sensitive` method.
3. Write a test for `as_case_sensitive` to ensure expected behavior.
4. Profit! These metadata queries should now take 100ms instead of 3s (see screenshot).

snowflake_query_runtimes

### Closed Questions

Let’s say I have 3 tables:
```
my_table
my_table_testing
my_table_backup
```

❓ Am I guaranteed that
`show objects in schema my_schema starts with 'my_table' limit 1` will always return metadata for `my_table` and never for the other two?
Since technically all three satisfy the `starts_with 'my_table'` condition.

**Answer from Snowflake: Yes, in this case, as you have specified limit 1, the output will be `my_table`.**

### Checklist

- [x] I have run this code in development, and it appears to resolve the stated issue
- [x] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX


mikealfare:
> Answer from Snowflake: Yes, in this case, as you have specified limit 1, the output will be my_table.

To be extra clear, this was Snowflake's response because `my_table` is an exact match, and not because it was the first in the example list, correct?

dbt/adapters/snowflake/relation.py
```diff
@@ -106,3 +106,17 @@ def dynamic_table_config_changeset(
...
+        for path in ComponentName:
+            if self.include_policy.get_part(path):
+                part = self.path.get_part(path)
+                if part:
+                    if self.quote_policy.get_part(path):
+                        path_part_map[path] = part
+                    else:
+                        path_part_map[path] = part.upper()
+
+        return self.replace_path(**path_part_map)
```
mikealfare, line 122: I think this is the case, but to confirm, `replace_path` will return a copy of `self` with the new attribution, it won't modify the relation in place, correct?

dbt/adapters/snowflake/relation.py
```diff
@@ -106,3 +106,17 @@ def dynamic_table_config_changeset(
...
             return config_change_collection
         return None
+
+    def as_case_sensitive(self) -> "SnowflakeRelation":
+        path_part_map = {}
+
+        for path in ComponentName:
+            if self.include_policy.get_part(path):
+                part = self.path.get_part(path)
+                if part:
```
mikealfare, line 116: Bonus style points if you use the walrus operator here.

dbt/adapters/snowflake/relation.py
```diff
@@ -106,3 +106,17 @@ def dynamic_table_config_changeset(
...
+                    if self.quote_policy.get_part(path):
```
mikealfare, line 117: Since it looks like Snowflake stores all of the metadata as `upper()` unless quoted, there's a part of me that wants to do this translation on the way in (e.g. inside `self.create()`) instead of each time it's needed. I think we could simplify logic elsewhere if we did that (e.g. `self._make_match_kwargs()`), and ultimately avoid needing this method.

dbt/adapters/snowflake/relation.py
aranke: Thanks, but I'm good 😆

aranke:
> To be extra clear, this was Snowflake's response because my_table is an exact match, and not because it was the first in the example list, correct?

I'm not sure, and I'm not sure if it matters since the objects are always returned in alphabetical order.

```sql
show objects in ANALYTICS_DEV.DBT_KARANKE starts with 'STG_HUBSPOT__CONTACT';
```

|created_on|name|database_name|schema_name|kind|comment|cluster_by|rows|bytes|owner|retention_time|owner_role_type|budget|is_dynamic|
|-|-|-|-|-|-|-|-|-|-|-|-|-|-|
|2024-07-08 08:07:40.690 -0700|STG_HUBSPOT__CONTACT|ANALYTICS_DEV|DBT_KARANKE|TABLE| | |639962|43180544|TRANSFORMER|1|ROLE| |N|
|2024-07-09 15:17:12.486 -0700|STG_HUBSPOT__CONTACT_FORM_SUBMISSIONS|ANALYTICS_DEV|DBT_KARANKE|VIEW| | |0|0|TRANSFORMER|1|ROLE| |N|
|2024-07-08 07:52:06.624 -0700|STG_HUBSPOT__CONTACT_TMP|ANALYTICS_DEV|DBT_KARANKE|VIEW| | |0|0|TRANSFORMER|1|ROLE| |N|

```sql
show objects in ANALYTICS_DEV.DBT_KARANKE starts with 'STG_HUBSPOT__CONTACT' LIMIT 1;
```

|created_on|name|database_name|schema_name|kind|comment|cluster_by|rows|bytes|owner|retention_time|owner_role_type|budget|is_dynamic|
|-|-|-|-|-|-|-|-|-|-|-|-|-|-|
|2024-07-08 08:07:40.690 -0700|STG_HUBSPOT__CONTACT|ANALYTICS_DEV|DBT_KARANKE|TABLE| | |639962|43180544|TRANSFORMER|1|ROLE| |N|

dbt/adapters/snowflake/relation.py
aranke: Created a follow-up issue to investigate: https://github.com/dbt-labs/dbt-snowflake/issues/1111

mikealfare: Thanks for creating the follow up issue. Per our conversation, this suggestion is out of scope for this PR and I'm comfortable moving forward with this as-is.

mikealfare:
> I'm not sure if it matters since the objects are always returned in alphabetical order.

Agreed, and this is the confirmation I was looking for.

mikealfare (approved):

PR #1109, opened 2024-07-09
Improve run times for large projects by reusing connections by default

### LINKED ISSUE #1082: [Feature] Improve key-pair auth performance

### Describe the feature

@VersusFacit's analysis:
```
Params:
Unix time command -- acceptable imprecision for the orders of magnitude we're dealing with here
5000 node project
average 2 runs each for each authentication method
dbt snowflake - user pass: 427.83s user 42.43s system 16% cpu 46:49.12 total
dbt snowflake - key pair: 1011.76s user 44.96s system 32% cpu 53:42.99 total
```
400 vs 1000 is quite a dramatic difference!

Avenues for investigation:
- We want to first try caching the keyfile contents
- Redo timing analysis to see if there's any benefit
- Look at any extra calls being made to snowflake
- If that fails, create a ticket for further investigation

mikealfare:
resolves #1082

### Problem

Run times for large projects are taking a long time, in part due to the fact that we don't reuse connections by default. This problem is made worse when using Key Pair authorization as the RSA key needs to be validated every time a connection is created.

### Solution

Update `reuse_connections` to a default of `True` if the user has not specified `client_session_keep_alive=True`.

### Checklist


tests/performance/test_auth_methods.py
```diff
@@ -0,0 +1,128 @@
...
+                "private_key_passphrase": os.getenv("SNOWFLAKE_PRIVATE_KEY_PASSPHRASE"),
+            }
+
+        else:
+            raise ValueError(
+                f"`auth_method` must be one of `user_password` or `key_pair`, received: {self.auth_method}"
+            )
+
+    def test_scenario(self, project):
+        run_dbt(["run"])
```
VersusFacit, line 104: So the goal here is to have a test that tracks metrics but no automated checks on those? More us keeping an eye on things?

tests/unit/test_snowflake_adapter.py
```diff
@@ -317,7 +317,7 @@ def test_client_session_keep_alive_true(self):
                     application="dbt",
                     insecure_mode=False,
                     session_parameters={},
-                    reuse_connections=None,
```
VersusFacit, line 320: Hm, so above I see that you set this logic to say "if reuse connections is None and client_session_keep_alive is True, leave it". Shouldn't this stay `None` then?

What's are these tests actually testing if not what configs are allowed / unallowed

VersusFacit (commented):
Judging from earlier conversations, does this PR need a rework? I think I saw the spec may have changed to default to `True` regardless here.

Slightly out of context, this is from you:
> we will update the default on reuse_connections to be True

Just want to know target

tests/performance/test_auth_methods.py
mikealfare: That's more or less correct. The test doesn't check if functionality is working or not working. It simply measures how well it's working relative to other methods on the specific machine where it's running; i.e. it always passes unless the connection is bad. If we were to throw an assertion in that ensured that it completed within a certain amount of time, then I'd refer to this more as a benchmarking test. But that opens up a whole new ball of wax, and we should discuss that before implementing something like that.

This makes it easier to re-run a performance analysis in the future. It also provides some quantitative context around the decisions made in this PR (turn on `reuse_connections` by default due to a large benefit, but don't turn on `unsafe_skip_rsa_key_validation` due to a negligible benefit), which may help if we were to ever look back at this PR.

mikealfare:
> Judging from earlier conversations, does this PR need a rework? I think I saw the spec may have changed to default to `True` regardless here.
> 
> Slightly out of context, this is from you:
> 
> > we will update the default on reuse_connections to be True
> 
> Just want to know target

Fair, I should have updated the PR body. After internal discussions, we determined that we only wanted to default `reuse_connections` to `True` if the user didn't specify `client_session_keep_alive=True` (and didn't set `reuse_connection`). We wanted to do that to make the potentially risky scenario (`reuse_connections=True` and `client_session_keep_alive=True`) an opt-in scenario.

tests/unit/test_snowflake_adapter.py
mikealfare: Yes, you're correct. I'm surprised this unit test passed, good catch.

mikealfare: Ah, it's because the `Credentials` class is initially created with `client_session_keep_alive=False`, hence the `__post_init__` sets the default for `reuse_connections`. Then `replace` is used to replace the property `client_session_keep_alive` which does not retrigger `__post_init__`.

mikealfare: I updated the test to set both, with an explanation as to why `reuse_connections` needs to be reset. I don't love it, but I don't know another way to test it without re-configuring this test class to not use `replace`. I'm open to suggestions.

tests/performance/test_auth_methods.py
```diff
@@ -0,0 +1,132 @@
+"""
+Results:
+
+| method        | project_size | reuse_connections | unsafe_skip_rsa_key_validation | duration |
+|---------------|--------------|-------------------|--------------------------------|----------|
+| User Password |        1,000 | False             | -                              |  234.09s |
+| User Password |        1,000 | True              | -                              |   78.34s |
+| Key Pair      |        1,000 | False             | False                          |  271.47s |
```
aranke, line 8: Have these stats been updated since yesterday’s key-pair caching PR?

If not, probably worth updating as a favor to our future selves 🙂

dbt/adapters/snowflake/connections.py
```diff
@@ -143,6 +144,12 @@ def __post_init__(self):
 
         self.account = self.account.replace("_", "-")
 
+        # only default `reuse_connections` to `True` if the user has not turned on `client_session_keep_alive`
+        # having both of these set to `True` could lead to hanging open connections, so it should be opt-in behavior
+        if self.client_session_keep_alive is False and self.reuse_connections is None:
+            print("Updating reuse_connections")
```
aranke, line 150: Do you want to keep the print here?
If you want to log something, look into `fire_event(Note(...))`.

tests/performance/test_auth_methods.py
mikealfare: Yeah, I reran them after the caching. I didn't rerun the user/password ones since they shouldn't be affected.

dbt/adapters/snowflake/connections.py
mikealfare: lol, no, good catch, I'll update it

PR #1110, opened 2024-07-10
Improve run times when using key pair auth by caching the private key

### LINKED ISSUE #1082: [Feature] Improve key-pair auth performance

mikealfare:
resolves #1082

### Problem

We evaluate and validate the private key on every connection when using key pair auth despite the fact that the key does not change during a dbt run.

### Solution

- create a function that uses caching outside of the `Credentials` class and pass the parameters to this function
- use `functools.lru_cache` if we're using Python 3.8
- use `functools.cache` if we're using Python 3.9
- remove the check on Python version when we retire Python 3.8 support

### Checklist


dbt/adapters/snowflake/connections.py
```diff
@@ -63,6 +72,73 @@
...
+        return None
+
+    return p_key.private_bytes(
+        encoding=serialization.Encoding.DER,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+
+
+@cache
```
aranke, line 99: Nit: `cache` here is redundant since this method should only get called exactly once.

dbt/adapters/snowflake/connections.py
```diff
@@ -273,44 +349,8 @@ def _get_access_token(self) -> str:
...
-        else:
-            return None
-
-        return p_key.private_bytes(
-            encoding=serialization.Encoding.DER,
-            format=serialization.PrivateFormat.PKCS8,
-            encryption_algorithm=serialization.NoEncryption(),
-        )
+    def _get_private_key(self) -> Optional[bytes]:
+        return _private_key(self.private_key, self.private_key_path, self.private_key_passphrase)
```
aranke, line 353: Nit: If you want to keep this method in the class, you can use `functools.cached_property`: https://docs.python.org/3/library/functools.html#functools.cached_property

aranke (commented):
Caching LGTM, with a couple of nits.
I'll let someone from Adapters provide a final review.

dbt/adapters/snowflake/connections.py
mikealfare: While it's highly unlikely, if someone builds on `dbt-snowflake` and overwrites this for some reason, we'll break them. Also, putting the body of the function outside of the class means I can test it without mocking anything. I'm looking at adding a test or two tonight.

dbt/adapters/snowflake/connections.py
mikealfare: Agreed, but I figured I might as well do it here too as it's part of the same logical collection.

tests/functional/auth_tests/test_key_pair.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+from dbt.tests.util import run_dbt
+import pytest
+
+
+class TestKeyPairAuth:
```
mikealfare, line 7: I have a feeling this test will fail because I doubt we have the private key in GHA.

mikealfare: The secrets are added and this is passing now.

dbt/adapters/snowflake/connections.py
```diff
@@ -63,6 +72,58 @@
 }
 
 
+@cache
```
colin-rogers-dbt, line 75: nit:  can we move these to their own module? Just want to keep connections.py small

mikealfare: I actually want to move them into `dbt-adapters` long term. I thought about creating a separate module, but thought it would be unnecessary if we were going to move it later. I can add that here though if you think it's worth it.

colin-rogers-dbt: I think it's worth it as it should make it cleaner to move later

mikealfare: I moved over only the two methods that are generic. I left the one that's Snowflake-specific (converting the key to DER format).

Along the same lines of making `connections.py` smaller, I actually think the `Credentials` class should be in its own module (probably named `auth`, even though that's what I moved these methods to). That's a separate PR, but it shouldn't take long to do if we want to do that.

colin-rogers-dbt (approved):

PR #1115, opened 2024-07-12
Python model logging

### LINKED ISSUE #846: [ADAP-1046] [CT-3418] [Bug] dbt python model logging not working with snowflake
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core

### Current Behavior

dbt python model logging statement not able to put logs inside snowflake event table.

Model Name: Logging_test.py
Code: 
```python
import snowflake.snowpark as snowpark
import snowflake.snowpark.functions as f
from snowflake.snowpark.functions import *
import logging
logger = logging.getLogger("dbt_logger")
logger.info("******Inside Logging module.******")
def model(dbt, session):
    session.sql(f"""ALTER SESSION SET LOG_LEVEL = INFO""").collect()
    logger.info("******Logging start.******")
    df=session.sql(f"""select current_user() as session_user,current_role() as session_role""")
    logger.info("******Logging End.******")
    return df
```

### Expected Behavior

dbt model should put log in event table 

### Steps To Reproduce

1. Set snowflake event table : https://docs.snowflake.com/en/developer-guide/logging-tracing/event-table-setting-up
2.  Set connection param in dbt profile.yml file
3.  run the dbt python model : dbt run -model logging_test
4. check event table in snowflake setup in step1 
    ```sql
    SELECT
	    RECORD['severity_text']::STRING log_level ,
	    VALUE::STRING message ,
	    RESOURCE_ATTRIBUTES['snow.query.id']::STRING query_id
    FROM
	    DXRX_OPERATIONS.LOGGING.EVENTS
    WHERE
	    SCOPE['name']::STRING = 'dbt_logger'
    ORDER BY
	    "TIMESTAMP" DESC;
    ```
5. Check query history in snowsight and get the dbt generated store procedure (Sql query start with  WITH logging_test__dbt_sp AS PROCEDURE ())
6. Execute the procedure from snowsight and wait for few minute
7. Execute the step 4 query again , it will show you logs generated by Snowsight query id 
8. dbt generated query id logs missing from event table ----(it should be there )

### Relevant log output

_No response_

### Environment

```markdown
- OS: window
- Python: 3.8
- dbt:1.7.2
```

snowflake

### Additional Context

While executing the dbt generated store procedure from snowsight, it's working fine and insert logs in event log table

mikealfare:
resolves #846

### Problem

We don't provide a standard way to enable logging for python models.

### Solution

Put together some proposals for minimal permission workarounds as well as some potential implementations for down the road. This likely won't be merged, but will serve as a PoC.

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -363,7 +363,6 @@ $$"""
         if use_anonymous_sproc:
             proc_name = f"{identifier}__dbt_sp"
             python_stored_procedure = f"""
-ALTER SESSION SET LOG_LEVEL = INFO;
 ALTER SESSION SET TRACE_LEVEL = ON_EVENT;
 WITH {proc_name} AS PROCEDURE ()
 {common_procedure_code}
@@ -372,8 +371,6 @@ CALL {proc_name}();
         else:
             proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
             python_stored_procedure = f"""
-ALTER SESSION SET LOG_LEVEL = INFO;
-ALTER SESSION SET TRACE_LEVEL = ON_EVENT;
 CREATE OR REPLACE PROCEDURE {proc_name} ()
 {common_procedure_code};
 CALL {proc_name}();
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -363,7 +363,6 @@ $$"""
         if use_anonymous_sproc:
             proc_name = f"{identifier}__dbt_sp"
             python_stored_procedure = f"""
-ALTER SESSION SET LOG_LEVEL = INFO;
 WITH {proc_name} AS PROCEDURE ()
 {common_procedure_code}
 CALL {proc_name}();
@@ -371,7 +370,6 @@ CALL {proc_name}();
         else:
             proc_name = f"{database}.{schema}.{identifier}__dbt_sp"
             python_stored_procedure = f"""
-ALTER SESSION SET LOG_LEVEL = INFO;
 CREATE OR REPLACE PROCEDURE {proc_name} ()
 {common_procedure_code};
 CALL {proc_name}();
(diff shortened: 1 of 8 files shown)


dbt/adapters/snowflake/impl.py
```diff
@@ -363,6 +363,8 @@ def submit_python_job(self, parsed_model: dict, compiled_code: str):
         if use_anonymous_sproc:
             proc_name = f"{identifier}__dbt_sp"
             python_stored_procedure = f"""
+ALTER SESSION SET LOG_LEVEL = DEBUG;
```
mikealfare, line 366: @manishkundu @wenbaolicat I think if you do this at the account level, you'll get the behavior you want. I know that's a high bar since you need account level access. But that should unblock you until we implement this at the "out-of-the-box" level.

tests/functional/adapter/python_model_tests/test_logging.py
```diff
@@ -0,0 +1,43 @@
...
+MODEL__LOGGING = """
+import logging
+
+import snowflake.snowpark as snowpark
+import snowflake.snowpark.functions as f
+from snowflake.snowpark.functions import *
+
+
+logger = logging.getLogger("dbt_logger")
+logger.setLevel(logging.INFO)
```
mikealfare, line 14: It looks like if we move the logging to use the Python `logging` module, then it works inside of the model, leaving only the `TRACE_LEVEL` setting as a sticking point.

dbt/adapters/snowflake/impl.py
mikealfare: Updated: Try running `ALTER ACCOUNT SET TRACE_LEVEL = ON_EVENT` before running your repro, and swap out the sql setting of `LOG_LEVEL` with the `logging` setting (`logger.setLevel(logging.INFO)`). It looks like this is working from my end, which means we may only need to implement the `TRACE_LEVEL` piece.

tests/functional/adapter/python_model_tests/test_logging.py
mikealfare: It looks like this might have been a timing issue. Apparently it takes over a minute for the events to show up in the event table, so this might have been associated with a previous scenario. It appears that setting the logging from within the python model does not work, whether running via SQL, `.setLevel()`, or setting `.level`.

dbt/adapters/snowflake/impl.py
mikealfare: Update 2, it looks like this was a timing issue and this does not work.

Issue #1123, opened 2024-07-18
[Bug] Concurrency issues in snowflake incremental models.


rattata2me:
### Is this a new bug in dbt-snowflake?

- [X] I believe this is a new bug in dbt-snowflake
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When running an incremental table with the on_schema_change policy set to append_new_columns, a race condition can occur. If two jobs perform the column check at the same time, the same ALTER TABLE statement will be generated, resulting in a SQL compilation error on the slower executor:
```
SQL Compilation error:
ERROR
  column '<new_column>' already exists
```
The core issue is that we cannot ensure the column schema remains unchanged between the column check operation and the ALTER TABLE statement.

### Expected Behavior

To avoid this issue, the process should either utilize the IF NOT EXISTS and IF EXISTS conditions provided by Snowflake or make use of transactions so that column checkup and update is atomic.

### Steps To Reproduce

1. Create an Incremental Table:
    Define and create an incremental table in your dbt project.
2. Add a Column:
    Add a new column entry to the model.
3. Concurrently Execute `dbt run`:
    Execute the dbt process in two separate jobs concurrently.

### Environment

```markdown
- OS:Ubuntu 22.04
- Python: 3.9
- dbt-core: 1.5.7
- dbt-snowflake: 1.5.7
```

amychen1776:
@rattata2me Could you say more about your use case here? Is there a reason why you would want to have concurrent execution against the same target (especially given that you are changing the target table schema with a new column).

rattata2me:
I have to dynamically update the table contents every time there is a new data ingestion, the data ingestion can be concurrent; coming from different users at the same time. This is a critical process which has to have 100% up time so I depend on the logic of on schema change to roll out column changes while keeping the service running.

rattata2me:
> @rattata2me Could you say more about your use case here? Is there a reason why you would want to have concurrent execution against the same target (especially given that you are changing the target table schema with a new column).

I have to dynamically update the table contents every time there is a new data ingestion, the data ingestion can be concurrent; coming from different users at the same time. This is a critical process which has to have 100% up time so I depend on the logic of on schema change to roll out column changes while keeping the service running.

amychen1776:
Thank you for the explanation!

PR #1132, opened 2024-07-19
Safe logic for concurrent executions

rattata2me:
resolves #
[docs](https://github.com/dbt-labs/dbt-snowflake/issues/1123)
### Problem
when running an incremental table with the on_schema_change policy set to append_new_columns in a dbt project. If two jobs concurrently perform the column check operation, they both generate the same ALTER TABLE statement. Because of this simultaneous execution, a race condition occurs where one job's ALTER TABLE statement succeeds, while the slower executor encounters a SQL compilation error stating that the column already exists. This issue stems from the inability to ensure that the column schema remains unchanged between the column check operation and the execution of the ALTER TABLE statement, leading to potential failures in concurrent environments.
### Solution
The solution to this problem is to incorporate the IF NOT EXISTS and IF EXISTS conditions in the ALTER TABLE statements. By using these conditions, the ALTER TABLE statement will only attempt to add a column if it does not already exist, and drop a column only if it exists.
However, this introduces some drawbacks. These include increased complexity in SQL logic, potential masking of underlying schema synchronization issues, minor performance impacts, and the risk of partial schema updates. Additionally, this solution is specific to Snowflake or databases that support these conditions, making it less portable to other database systems.
### Checklist
- [ X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ X] I have run this code in development and it appears to resolve the stated issue
- [ X] This PR includes tests, or tests are not required/relevant for this PR
- [ X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX


rattata2me:
This is my proposed solution to resolve the stated concurrency issues stated in https://github.com/dbt-labs/dbt-snowflake/issues/1123 . Right now I can not think of any major drawbacks of fixing the issue by usign the 'IF EXISTS' statements, any feed back on this is appreciated.

colin-rogers-dbt:
By 'concurrent executions' do you mean performing dbt commands like `dbt run`  against the same target in parallel?

rattata2me:
Yes precisely, this happens when executing `dbt run` in two separate processes at the same time against the same target table.

amychen1776:
At this time - we do not intend to support concurrent dbt runs and thus will be closing this PR for now.

PR #1149, opened 2024-07-25
Integration workflow update to support all-in-one adapter testing

gshank:
resolves #

### Problem

When testing changes to adapters, particularly dbt-adapters changes but also some core changes, we need to execute integration tests on all of the adapters, with the ability to specify dbt-adapters branch, dbt-core branch, and sometimes the local adapter branch. 

### Solution

Add additional input fields for the "workflow_dispatch" event: "dbt_adapter_branch", and "name".   The name helps to identify individual runs in the lists of workflow runs, which can get lost. Also there's no good way to connect up a rest api workflow run or a gh cli workflow run with the matching run result, so adding a distinctive name assists in finding the correct results.

### Checklist


.github/workflows/integration.yml
```diff
@@ -34,10 +36,26 @@ on:
   # manual trigger
   workflow_dispatch:
```
colin-rogers-dbt, line 37: Do we anticipate other workflows (i.e. in dbt-core) wanting to run this workflow? Should we also add `workflow_call`?

gshank: Workflow_call just calls the action code, without actually executing in this repository. We might want to standardize on an action in a central repository, but I'm guessing it wouldn't be in dbt-snowflake. Workflow_dispatch works for cross-repo calls. I'm running these now with a script, but once the individual repository changes are in place I'm going to try to set up an action in dbt-adapters to call all of the individual adapter integration tests and see how that goes.

.github/scripts/update_dev_packages.sh
mikealfare, line 1: I would leave "branch" in the name (e.g. `update_dev_dependency_branches.sh`). Using "packages" here makes it seem like we're updating or upgrading packages that are already installed.

.github/workflows/integration.yml
```diff
@@ -20,6 +20,8 @@
 
 name: Adapter Integration Tests
 
+run-name: "${{ contains(github.event_name, 'workflow_') && inputs.name || github.event_name }}: ${{ contains(github.event_name, 'workflow_') &&  inputs.adapter_branch || github.ref_name }} by @${{ github.actor }}"
```
mikealfare, line 23: I know how the mechanic `contains() && string` works, but it's not intuitive. Between that and the combination of an `&&` and an `||`, I would put parentheses around the `&&` expressions.

.github/workflows/integration.yml
```diff
@@ -34,10 +36,26 @@ on:
...
+        description: "name to associate with run"
+        required: false
+        type: string
+        default: "Adapter Integration Tests"
+      dbt_adapters_branch:
+        description: "The branch of dbt-adapters to use"
+        type: string
+        required: false
+        default: main
+      adapter_branch:
```
mikealfare, line 54: I would call this `dbt_snowflake_branch` to align with `dbt_adapters_branch` and `dbt_core_branch`. We already need to know we're calling a workflow in the `dbt-snowflake` repo, so we don't gain much from making it generic.

.github/workflows/integration.yml
```diff
@@ -34,10 +36,26 @@ on:
   # manual trigger
   workflow_dispatch:
     inputs:
```
mikealfare, line 38: The ordering of `dbt_core_branch`, `name`, `dbt_adapters_branch`, `adapter_branch` feels odd. I would group the branches either in descending or ascending order (e.g. `dbt_core_branch`, `dbt_snowflake_branch`, `dbt_adapters_branch`) and then put `name` either before or after that group. I would also make the arguments that go into the shell script are in the same order as the inputs to this workflow.

I think we might as well add `dbt-common` to this list as well. I know it's rare that we need to test against a feature branch of `dbt-common`, but that's what defaults are for.

.github/workflows/integration.yml
```diff
@@ -34,10 +36,26 @@ on:
...
-      dbt-core-branch:
+      dbt_core_branch:
         description: "branch of dbt-core to use in dev-requirements.txt"
         required: false
         type: string
+        default: main
+      name:
```
mikealfare, line 44: What's the value of this parameter. Is it really just to insert into `run-name` to identify the dispatched run? I feel like `github.event_name` and `github.actor` would suffice. Removing this would simplify the interface.

.github/workflows/integration.yml
```diff
@@ -141,12 +159,19 @@ jobs:
 
     steps:
       - name: Check out the repository
-        if: github.event_name != 'pull_request_target'
+        if: github.event_name != 'pull_request_target' && github.event_name != 'workflow_dispatch'
```
mikealfare, line 162: Since there are only three options, can we update this conditional to be stated in the positive, i.e. `if: github.event_name == 'push'`?

mikealfare (commented):
Mostly nits, but I do think it would make it smoother. Let me know what you think.

.github/scripts/update_dev_packages.sh
gshank: I used that name because that's what the script was already named in dbt-postgres. I agree that it's not particularly meaningful so I'll change the names.

.github/workflows/integration.yml
gshank: The exact same call is going to be made to each repo, so giving it the same field name DOES simplify it a lot. If every repo has a different field name for the local branch, then a different call has to be constructed for every repo.

.github/workflows/integration.yml
gshank: I was wondering about dbt_common... It is less common but you're right that we might as well handle it all.

.github/workflows/integration.yml
gshank: I want to be able to query all of the runs that I've created for a particular pull request, and event_name and actor are not enough to determine that. If github's apis returned the run id like it should this would be moot, but it doesn't. The only way that I found to find the run that you dispatched is by matching on a unique name.  Right now I'm working on FOUR different pull requests that need this testing and if all I have is event_name and actor they'll all get mixed up.

.github/workflows/integration.yml
gshank: We could. The problem is that right now different adapter repo versions of this file have a different set of event_names they'll process, so I'll have to read each one carefully to make sure that everything is handled and since I imagine that at some point we'd like to have a common version instead of 5 subtly different versions, it felt more robust to name each one. It's certainly a judgment call.

gshank: Also if somebody adds an additional event_name, it wouldn't default then and they'd have to remember to update these sections.

.github/workflows/integration.yml
mikealfare: Agreed, `workflow_dispatch` is effectively `workflow_call` + the workflow repo's code in this scenario, which is what we want.

.github/workflows/integration.yml
```diff
@@ -34,10 +36,31 @@ on:
...
+      name:
+        description: "Name to associate with run (example: 'dbt-adapters-242')"
         required: false
         type: string
+        default: "Adapter Integration Tests"
+      adapter_branch:
+        description: "The branch of this adapter repository to use"
+        type: string
+        required: false
+        default: main
```
mikealfare, line 48: Minor nit, but the missing double quotes is making my OCD twitch. Alternatively, remove double quotes from the other string values, which appears to be the style in this particular file (e.g. all of the existing name fields don't use quotes).

.github/scripts/update_dev_packages.sh
gshank: Renamed.

.github/workflows/integration.yml
gshank: Parentheses added.

.github/workflows/integration.yml
gshank: dbt-common added. Rearranged order of fields to "name", "adapter_branch", "dbt_adapters_branch", "dbt_core_branch", and "dbt_common_branch", because that feels like the natural order to me.

gshank:
Pull requests in other repos:

https://github.com/dbt-labs/dbt-postgres/pull/133

https://github.com/dbt-labs/dbt-bigquery/pull/1309

https://github.com/dbt-labs/dbt-redshift/pull/889

https://github.com/dbt-labs/dbt-spark/pull/1082

.github/workflows/integration.yml
gshank: Fixed.

.github/workflows/integration.yml
gshank: You've convinced me. I've changed them to `== 'push'`.

mikealfare (approved):

PR #1161, opened 2024-08-07
Isolating distribution testing

### LINKED ISSUE #1130: Update the build check workflow
### Current Behavior

We install the wheel, check that the package is installed, then immediately install the sdist, and check that the package is installed again.

### Expected Behavior

We should be validating the packages in isolated environments. This arose from the `setuptools` / `packaging` issue for `dbt-core<1.4`, but it's something we should do for current versions as well. See https://github.com/dbt-labs/dbt-core/issues/10465 for the analogous `dbt-core` issue.

leahwicz:
resolves #1130 

### Problem
We validate wheel and source distributions in the same environment during the build validation step. This led to issues documented here https://github.com/dbt-labs/dbt-core/issues/10465. 

### Solution
This splits out the wheel and source installation and testing into their own Action's job to test

This is a test run of the Action running with the installation and testing of wheels and source distributions split into separate Actions Jobs that isolate them from one another: https://github.com/leahwicz/dbt-snowflake/actions/runs/10276531967/job/28437083967 

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -163,9 +163,9 @@ jobs:
           overwrite: true
 
   test-build:
-    name: verify packages / python ${{ matrix.python-version }} / ${{ matrix.os }}
+    name: verify packages / python ${{ matrix.python-version }} / ${{ matrix.os }} / ${{ matrix.dist-type }} 
 
-    if: needs.build.outputs.is_alpha == 0
+    # if: needs.build.outputs.is_alpha == 0
 
     needs: build
 
@@ -176,6 +176,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-12, windows-latest]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        dist-type: ['whl', 'gz']
 
     steps:
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


mikealfare (approved):

.github/workflows/main.yml
```diff
@@ -163,7 +163,7 @@ jobs:
           overwrite: true
 
   test-build:
-    name: verify packages / python ${{ matrix.python-version }} / ${{ matrix.os }}
+    name: verify packages / python ${{ matrix.python-version }} / ${{ matrix.os }} / ${{ matrix.dist-type }}
```
mikealfare, line 166: I think you're getting dinged for the extra space at the end of this line. Try running `precommit`.

leahwicz: I promise I'll run this

mikealfare:
@leahwicz Nice try on re-running the tests. The problem is that the change is only a workflow file and we [filter](https://github.com/dbt-labs/dbt-snowflake/blob/7fb4549abc52d53dfdbf02da423ee58cc8696ccc/.github/workflows/integration.yml#L118) on those for running this test. An admin needs to override it. Alternatively, we should probably just remove that from the filter since it's completely reasonable for a PR to only be workflow updates.

Issue #321, opened 2022-11-10
[CT-1492] [Feature] Add support for Iceberg Tables


teej:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-snowflake functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Snowflake has announced support for [Iceberg Tables](https://www.snowflake.com/blog/iceberg-tables-powering-open-standards-with-snowflake-innovations/), an external table format but with full interoperability as a 1st class snowflake table.

Once this feature goes live, dbt-snowflake should support materializing a model as an external iceberg table.

### Who will this benefit?

- dbt users who want their models available as parquet files in storage they control
- dbt users who want an easy way to interoperate with non-dbt systems that use the iceberg table format
- dbt users who want direct access and control over the underlying storage of their data

### Are you interested in contributing this feature?

yes

### Anything else?

Iceberg Tables are in private preview as of 2022-11-10

jaklan:
Any chance to put more light on that? We really need a Snowflake equivalent of https://github.com/dbt-athena/dbt-athena to push forward the adoption of open table format (which, I believe, would be also pretty crucial for `dbt` to support in the long term).

multiprocessingh:
+1 here! This would be super useful to support for enterprise sas use cases!

WangCHEN9:
+1 here!

DLewisEndeavor:
+1 here! Looking into Iceberg for enterprise sharing across platforms

apd-bscally:
+1, Iceberg is now in public preview on Snowflake

mourtallah:
+1 here!

VSchmidts:
+1 here!

ksdaly:
+1

sfc-gh-ripu:
+1

jimvekemans:
+1, will work on this a bit in the coming weeks and, if successful, open a pull request

dennysrega1ado:
+1

waitsfornone:
+1 here. We are adopting Iceberg tables as our data lake format and I would love to generate the tables out of Snowflake with dbt.

mikezusev:
+1

zewang-sanofi:
+1

[...]

PR #1170, opened 2024-08-22
Add support for Iceberg Table Materialization

VersusFacit:
related to #321 

# Add support for Iceberg Table Materialization

Built atop jaffle shop classic for testing.

## Problem

We want to add support for materialization S3 bucket Iceberg tables in Snowflake.

Example of our decided model config interface:
```sql
{{
  config(
    materialized = "table",
    table_format="iceberg",
    external_volume="s3_iceberg_snow",
    base_location="milas_working_folder",
  )
}}
-- transient gets auto-`falsed`
-- `transient=true` is ignored
select * from {{ ref('raw_orders') }}
```

## Solution

* Precisely insert additional Iceberg DDL into existing `table` materialization
* load all custom logic possible into Python
* manually attach `is_iceberg` as a field to `show objects` (this will be given to us by Snowflake for free one day)
* use `is_iceberg` to determine `drop_relation_if_exists` calls before building a table (table -> Iceberg, Iceberg -> table)
* gatekeep behind a behavior flag for those who have no need of Iceberg format tables!

## Caveats
this is ready for review but needs two more things I can identify:
1. a behavior flag to gate Iceberg (did we merge that capability yet?)
2. activate tests (but while we have an Iceberg volume on our team cluster, it doesn't seem it's live yet on the CI one here)

## Impact
Get this shipped for Cloud folks to start using!!

### Manually testing scenarios
* [x] build an Iceberg table that refs a table
* [x] build an iceberg table on top of a table
* [x] seamless `create`s from table to iceberg table and back
* [x] did stress tests to confirm memory usage / time tradeoff of `left join` with `show objects` was worth it
* [x] a complex config model, with cluster by and iceberg at the same time

### Checklist

- [x] ~~(well, err, sort of)~~ edit: Now, this PR includes tests, or tests are not required/relevant for this PR 

### DIFF
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -44,6 +44,11 @@ class SnowflakeConfig(AdapterConfig):
     merge_update_columns: Optional[str] = None
     target_lag: Optional[str] = None
 
+    # extended formats
+    object_format: Optional[str] = None
+    external_volume: Optional[str] = None
+    base_location: Optional[str] = None
+
 
 class SnowflakeAdapter(SQLAdapter):
     Relation = SnowflakeRelation
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -9,6 +9,7 @@ from dbt.adapters.sql.impl import (
     LIST_SCHEMAS_MACRO_NAME,
     LIST_RELATIONS_MACRO_NAME,
 )
+from dbt_common.behavior_flags import BehaviorFlag
 from dbt_common.contracts.constraints import ConstraintType
 from dbt_common.contracts.metadata import (
     TableMetadata,
@@ -76,6 +77,10 @@ class SnowflakeAdapter(SQLAdapter):
             Capability.GetCatalogForSingleRelation: CapabilitySupport(support=Support.Full),
         }
     )
+
+    @property
+    def _behavior_flags(self) -> List[BehaviorFlag]:
+        return [{"name": "enable_iceberg_materializations", "default": False}]
 
     @classmethod
     def date_function(cls):
@@ -253,12 +258,17 @@ class SnowflakeAdapter(SQLAdapter):
 
     def _parse_list_relations_result(self, result: "agate.Row") -> SnowflakeRelation:
         # this can be reduced to always including `is_dynamic` once bundle `2024_03` is mandatory
+        # this can be reduced to always including `is_iceberg` once Snowflake adds it to show objects
         try:
-            database, schema, identifier, relation_type, is_dynamic, is_iceberg = result
+            if self.behavior.enable_iceberg_materializations.no_warn:
+                database, schema, identifier, relation_type, is_dynamic = result
+            else:
+                database, schema, identifier, relation_type, is_dynamic, is_iceberg = result
         except ValueError:
             database, schema, identifier, relation_type = result
             is_dynamic = "N"
-            is_iceberg = "N"
+            if self.behavior.enable_iceberg_materializations.no_warn:
+                is_iceberg = "N"
 
         try:
             relation_type = self.Relation.get_relation_type(relation_type.lower())
@@ -268,8 +278,12 @@ class SnowflakeAdapter(SQLAdapter):
         if relation_type == self.Relation.Table and is_dynamic == "Y":
             relation_type = self.Relation.DynamicTable
 
+        # This line is the main gate on supporting Iceberg materializations. Pass forward a default
+        # table format, and no downstream table macros can build iceberg relations.
         table_format: str = (
-            TableFormat.ICEBERG if is_iceberg in ("Y", "YES") else TableFormat.DEFAULT
+            TableFormat.ICEBERG
+            if self.behavior.enable_iceberg_materializations.no_warn and is_iceberg in ("Y", "YES")
+            else TableFormat.DEFAULT
         )
         quote_policy = {"database": True, "schema": True, "identifier": True}
 
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -233,34 +233,8 @@ class SnowflakeAdapter(SQLAdapter):
     ) -> List[SnowflakeRelation]:
         kwargs = {"schema_relation": schema_relation}
 
-        def check_is_iceberg(row, table2):
-            for match_row in table2.rows:
-                if (
-                    row["name"] == match_row["name"]
-                    and row["database_name"] == match_row["database_name"]
-                    and row["schema_name"] == match_row["schema_name"]
-                ):
-                    return "Y"
-            return "N"
-
         try:
             schema_objects = self.execute_macro(LIST_RELATIONS_MACRO_NAME, kwargs=kwargs)
-            iceberg_table_results = self.execute_macro(
-                LIST_ICEBERG_RELATIONS_MACRO_NAME, kwargs=kwargs
-            )
-            import agate
-
-            # this only seems to only inflate runtime 16%; TODO: stress test
-            results = schema_objects.compute(
-                [
-                    (
-                        "is_iceberg",
-                        agate.Formula(
-                            agate.Text(), lambda row: check_is_iceberg(row, iceberg_table_results)
-                        ),
-                    )
-                ]
-            )
         except DbtDatabaseError as exc:
             # if the schema doesn't exist, we just want to return.
             # Alternatively, we could query the list of schemas before we start
@@ -271,12 +245,12 @@ class SnowflakeAdapter(SQLAdapter):
 
         # this can be reduced to always including `is_dynamic` once bundle `2024_03` is mandatory
         columns = ["database_name", "schema_name", "name", "kind"]
-        if "is_dynamic" in results.column_names:
+        if "is_dynamic" in schema_objects.column_names:
             columns.append("is_dynamic")
-        if "is_iceberg" in results.column_names:
+        if "is_iceberg" in schema_objects.column_names:
             columns.append("is_iceberg")
 
-        return [self._parse_list_relations_result(result) for result in results.select(columns)]
+        return [self._parse_list_relations_result(obj) for obj in schema_objects.select(columns)]
 
     def _parse_list_relations_result(self, result: "agate.Row") -> SnowflakeRelation:
         # this can be reduced to always including `is_dynamic` once bundle `2024_03` is mandatory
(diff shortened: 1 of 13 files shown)


dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -44,13 +71,16 @@
...
+        alter {{ alter_statement_format_prefix }} table {{relation}} cluster by ({{cluster_by_string}});
       {%- endif -%}
       {% if enable_automatic_clustering and cluster_by_string is not none and not temporary  -%}
-        alter table {{relation}} resume recluster;
+        alter {{ alter_statement_format_prefix }} table {{relation}} resume recluster;
       {%- endif -%}
 
   {%- elif language == 'python' -%}
+    {%- if iceberg -%}
+      {% do exceptions.raise_compiler_error('Iceberg is incompatible with Python models. Please use a SQL model for the iceberg format.') %}
```
VersusFacit, line 82: I actually dont know if this is true

dbt/adapters/snowflake/impl.py
```diff
@@ -223,8 +232,35 @@ def list_relations_without_caching(
         self, schema_relation: SnowflakeRelation
     ) -> List[SnowflakeRelation]:
         kwargs = {"schema_relation": schema_relation}
+
+        def check_is_iceberg(row, table2):
```
VersusFacit, line 236: My day is ruined. Let us figure out a better option for this.

VersusFacit: We're going to talk to Snowflake about metadata

tests/functional/adapter/list_relations_tests/test_show_objects.py
```diff
@@ -73,7 +73,7 @@ def list_relations_without_caching(project) -> List[SnowflakeRelation]:
             database=project.database, schema=project.test_schema, identifier=""
         )
         with get_connection(my_adapter):
-            relations = my_adapter.list_relations_without_caching(schema)
+            relations = my_adapter.list_relations_without_caching(schema.path.schema)
```
VersusFacit, line 76: The original value here was `DATABASE.SCHEMA.` so I just used the `SCHEMA` string itself

dbt/include/snowflake/macros/relations/table/create.sql
VersusFacit: It is for now!

dbt/adapters/snowflake/relation.py
```diff
@@ -1,30 +1,41 @@
...
     SnowflakeDynamicTableRefreshModeConfigChange,
     SnowflakeDynamicTableTargetLagConfigChange,
     SnowflakeDynamicTableWarehouseConfigChange,
+    SnowflakeObjectFormat,
     SnowflakeQuotePolicy,
     SnowflakeRelationType,
 )
 
+if TYPE_CHECKING:
+    from dbt.artifacts.resources.v1.model import ModelConfig
```
VersusFacit, line 31: Hopefully I'm being clever enough here. If I'm not, happy to remove this and keep the type hints as strings.

colin-rogers-dbt: Instead of using the ModelConfig from dbt-core we should rely on the RelationConfig protocol: https://github.com/dbt-labs/dbt-adapters/blob/main/dbt/adapters/contracts/relation.py#L49

VersusFacit: Done and done. Thanks for the pointer!

dbt/adapters/snowflake/relation.py
```diff
@@ -120,3 +131,68 @@ def as_case_sensitive(self) -> "SnowflakeRelation":
                         path_part_map[path] = part.upper()
 
         return self.replace_path(**path_part_map)
+
+    def get_ddl_prefix_for_create(self, config: RelationConfig, temporary: bool):
```
colin-rogers-dbt, line 135: Nice, this looks like it will be easy to unit test

dbt/include/snowflake/macros/materializations/table.sql
```diff
@@ -85,3 +84,34 @@ def main(session):
 # dbt = dbtObj(session.table)
 # df = model(dbt, session)
 {%endmacro%}
+
+
+{% macro drop_old_relation_as_needed(old_relation, target_relation) %}
```
colin-rogers-dbt, line 89: Could we move this logic into python? It's probably easier as we could write this comparison logic into the Relation class

dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -26,7 +20,16 @@
 
     {{ sql_header if sql_header is not none }}
 
-        create or replace {{ table_type }} table {{ relation }}
+        create or replace {{ materialization_prefix }} table {{ relation }}
+        {%- if relation.is_iceberg_format %}
+          {#
+            Valid DDL in CTAS statements. Plain create statements have a different order.
```
mikealfare, line 26: The order of the options matters? That's wild. Also, I think you can drop the `{% else %}` below since there's nothing in the block.

tests/functional/iceberg/test_table_basic.py
```diff
@@ -0,0 +1,98 @@
...
+
+_MODEL_BASIC_TABLE_MODEL = """
+{{
+  config(
+    materialized = "table",
+    cluster_by=['id'],
+  )
+}}
+select 1 as id
+""".strip()
```
mikealfare, line 15: The only reason I've seen to use `.strip()` in these fixtures is for inline CSVs; the extra return gets read as an empty record. It doesn't hurt for models, but it's also not necessary.

tests/functional/iceberg/test_table_basic.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class TestIcebergTableTypeBuildsOnExistingTable:
+    model_name = "my_model.sql"
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {self.model_name: _MODEL_TABLE_FOR_SWAP}
+
+    def test_changing_model_types(self, project):
```
mikealfare, line 84: Can we add a view to this swapping test too? I think it would work since we probably just check if it's not a view and drop it, but it was scenarios like that which caught me on DTs and MVs.

tests/functional/adapter/list_relations_tests/test_show_objects.py
mikealfare: Was this causing test failures? The type on this argument is `SnowflakeRelation`, not `str`, in `SnowflakeAdapter`.

dbt/include/snowflake/macros/materializations/table.sql
mikealfare: Agreed, something like this would read nicely:

```sql
{% if target_relation.needs_to_drop(old_relation) %}
    {{ drop_relation_if_exists(old_relation) }}
{% endif %}
```

dbt/adapters/snowflake/impl.py
```diff
@@ -29,6 +32,7 @@
     import agate
 
 SHOW_OBJECT_METADATA_MACRO_NAME = "snowflake__show_object_metadata"
+LIST_ICEBERG_RELATIONS_MACRO_NAME = "snowflake__show_iceberg_relations"
```
mikealfare, line 35: I don't see where this gets used.

dbt/adapters/snowflake/relation.py
```diff
@@ -120,3 +131,68 @@ def as_case_sensitive(self) -> "SnowflakeRelation":
...
+            return ""
+
+    def get_ddl_prefix_for_alter(self):
+        """All ALTER statements on Iceberg tables require an ICEBERG prefix"""
+        if self.is_iceberg_format:
+            return "iceberg"
+        else:
+            return ""
+
+    def render_iceberg_ddl(self, config: RelationConfig):
```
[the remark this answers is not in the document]
colin-rogers-dbt: I think `get_iceberg_ddl_options` might be more consistent with the other methods here

tests/functional/adapter/list_relations_tests/test_show_objects.py
VersusFacit: It was causing errors.

dbt/adapters/snowflake/relation.py
VersusFacit: I'd rather not shorten the names. Let's have them actually what they are doing. 

Also whoops, of course I can add return types!

tests/functional/iceberg/test_table_basic.py
VersusFacit: You got it. Added!

tests/functional/iceberg/test_table_basic.py
VersusFacit: Oh sure, force of habit at work. Removed.

dbt/include/snowflake/macros/relations/table/create.sql
VersusFacit: yeah there's an alternate CREATE TABLE where the params are in a different place entirely.

And yep that's a leftover artifact. Removed!

dbt/adapters/snowflake/relation.py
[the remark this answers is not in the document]
VersusFacit: That's fair! Renamed.

I want to keep it in Python in anticipation of other Iceberg options we'll need to add!

dbt/include/snowflake/macros/materializations/table.sql
VersusFacit: Thirded, this is great!

dbt/adapters/snowflake/impl.py
VersusFacit: Leftover from earlier work. Cleaned!

dbt/adapters/snowflake/relation_configs/formats.py
```diff
@@ -0,0 +1,9 @@
+from dbt_common.dataclass_schema import StrEnum  # doesn't exist in standard library until py3.11
+
+
+class SnowflakeObjectFormat(StrEnum):
```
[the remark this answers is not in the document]
VersusFacit: That was the tension i was thinking. Based on all our conversations, I'll call it `TableFormat`, but note in docstrings that Snowflake calls this an object format.

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -139,11 +139,23 @@
   {%- set max_total_results = max_results_per_iter * max_iter -%}
   {% if schema_relation is string %}
     {%- set sql -%}
-      show objects in {{ schema_relation }} limit {{ max_results_per_iter }}
+      show objects in {{ schema_relation }} limit {{ max_results_per_iter }};
+      select all_objects.*, is_iceberg as "is_iceberg"
```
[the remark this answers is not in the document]
VersusFacit: Oh of course! Just turn it inside a bit.

tests/functional/adapter/list_relations_tests/test_show_objects.py
VersusFacit: Well I reverted this because it's working again. I must have introduced a bug way up the chain.

VersusFacit:
Blocked by behavior flag / Jinja security issue. Once we have a solution forward, this PR is ready for code review round 2!

colin-rogers-dbt (approved):

PR #1183, opened 2024-09-17
Add support for Iceberg table format in Dynamic Tables

mikealfare:
resolves #

### Problem

`dbt-snowflake` does not currently support Iceberg table format when using dynamic tables.

### Solution

- catch the user configuration required for Iceberg support
- update dynamic table relation templates to include the new config
- put this new config behind the Iceberg behavior flag to make this feature opt-in

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240917-100505.yaml
+++ b/.changes/unreleased/Features-20240917-100505.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add support for Iceberg table format in Dynamic ta
+time: 2024-09-17T10:05:05.609859-04:00
+custom:
+  Author: mikealfare
+  Issue: "123"


dbt/include/snowflake/macros/relations/dynamic_table/replace.sql
```diff
@@ -1,18 +1,84 @@
 {% macro snowflake__get_replace_dynamic_table_sql(relation, sql) -%}
+{#-
```
mikealfare, line 2: These doc blocks could be helpful. They also add a ton of lines to the file and they're not as easy to read and mentally separate from the functional code as they would be in python. I could take them or leave them. I do want to keep the urls referenced below though, since these macros mirror those docs very closely.

dbt/adapters/snowflake/impl.py
```diff
@@ -83,7 +83,7 @@ def _behavior_flags(self) -> List[BehaviorFlag]:
         return [
             {
                 "name": "enable_iceberg_materializations",
-                "default": False,
+                "default": True,
```
VersusFacit, line 86: Will this be turned back off before merge?

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -55,11 +60,12 @@ class SnowflakeDynamicTableConfig(SnowflakeRelationConfigBase):
     query: str
     target_lag: str
     snowflake_warehouse: str
+    catalog: SnowflakeCatalogConfig
     refresh_mode: Optional[RefreshMode] = RefreshMode.default()
     initialize: Optional[Initialize] = Initialize.default()
 
     @classmethod
-    def from_dict(cls, config_dict) -> "SnowflakeDynamicTableConfig":
+    def from_dict(cls, config_dict: Dict[str, Any]) -> Self:
```
VersusFacit, line 68: I tried adding Self but it said this wasn't allowed for our current python version!? Maybe because you're using `typing_extensions` instead of `typing`?

dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -69,12 +75,12 @@ def from_dict(cls, config_dict) -> "SnowflakeDynamicTableConfig":
             "query": config_dict.get("query"),
             "target_lag": config_dict.get("target_lag"),
             "snowflake_warehouse": config_dict.get("snowflake_warehouse"),
+            "catalog": SnowflakeCatalogConfig.from_dict(config_dict["catalog"]),
```
VersusFacit, line 78: Why not get? Is this guaranteed to exist?

dbt/include/snowflake/macros/relations/dynamic_table/replace.sql
VersusFacit: I like the doc blocks but I agree they are harder to read unfortunately. Worth the tradeoff of including them for now imo.

One other thing is to have them in Jinja {# #} then prefacing each line with -- to make them read as SQL comments, which helps.

VersusFacit (commented):
Preliminary comments

dbt/adapters/snowflake/impl.py
mikealfare: Yeah, that's how I was testing that things worked instead of duplicating all the tests. Now that I streamlined the tests, I will update them to use user overrides instead. Before I would have had to duplicate ~20 tests.

dbt/include/snowflake/macros/relations/dynamic_table/replace.sql
mikealfare: > One other thing is to have them in Jinja {# #} then prefacing each line with -- to make them read as SQL comments, which helps.

Good idea. I'll update it. I thought about how docstrings are easier to read in python because the IDE formats it differently. I didn't think to do that though.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: > Maybe because you're using typing_extensions instead of typing

Correct, `Self` isn't in `typing` until something like 3.11.

dbt/adapters/snowflake/relation_configs/dynamic_table.py
mikealfare: Yes, this is guaranteed to exist. Even if Iceberg is off, this defaults to the default table type on the catalog object. The describe query returns None for the entire catalog object, which is checked in the catalog parsing method. It's a bit clunky; the jinja template should really return 0 rows instead of a None object. But I'd rather put the logic in python than jinja.

tests/functional/relation_tests/dynamic_table_tests/models.py
```diff
@@ -48,3 +62,31 @@
...
+
+
+DYNAMIC_ICEBERG_TABLE_ALTER = """
+{{ config(
+    materialized='dynamic_table',
+    snowflake_warehouse='DBT_TESTING',
+    target_lag='5 minutes',
+    refresh_mode='INCREMENTAL',
+    table_format="iceberg",
+    external_volume="s3_iceberg_snow",
```
colin-rogers-dbt, line 74: Can we add a test to validate changing the external volume works?

colin-rogers-dbt: (realizing that means we need to setup a second external volume)

dbt/adapters/snowflake/relation_configs/catalog.py
```diff
@@ -0,0 +1,121 @@
...
+        if relation_config.config.extra.get("table_format") is None:
+            return {"table_format": "default"}
+
+        config_dict = {
+            "table_format": relation_config.config.extra.get("table_format"),
+            "name": "SNOWFLAKE",  # this is not currently configurable
+        }
+
+        if external_volume := relation_config.config.extra.get("external_volume"):
+            config_dict.update({"external_volume": external_volume})
```
VersusFacit, line 83: Why not just `config_dict["external_volume"]` since keys auto-add themselves and that's very pythonic.

dbt/adapters/snowflake/relation_configs/catalog.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if base_location := relation_config.config.extra.get("base_location_subpath"):
+            config_dict.update({"base_location": base_location})
+
+        return config_dict
+
+    @classmethod
+    def parse_relation_results(cls, relation_results: RelationResults) -> Dict[str, Any]:
+        catalog_results: "agate.Table" = relation_results["catalog"]
```
VersusFacit, line 92: Is this a safe access on agate.Table? It wouldn't be on a normal dictionary, but I assume this is just how Agate does things

dbt/adapters/snowflake/relation_configs/catalog.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if catalog_results is None or len(catalog_results) == 0:
```
VersusFacit, line 94: Python tends to like 

`if not catalog_results` since empty collections are falsey

over

`if catalog_results is None or len(catalog_results) == 0:`

Are the rules different for Agate tables?

dbt/adapters/snowflake/relation_configs/catalog.py
```diff
@@ -0,0 +1,121 @@
...
+            return {"table_format": "default"}
+
+        # for now, if we get catalog results, it's because this is an iceberg table
+        # this is because we only run `show iceberg tables` to get catalog metadata
+        # this will need to be updated once this is in `show objects`
+        catalog: "agate.Row" = catalog_results.rows[0]
+        config_dict = {"table_format": "iceberg"}
+
+        if name := catalog.get("catalog_name"):
+            config_dict.update({"name": name})
```
VersusFacit, line 104: once more regarding earlier syntax suggestion

dbt/adapters/snowflake/relation_configs/catalog.py
```diff
@@ -0,0 +1,121 @@
...
+        return config_dict
+
+
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class SnowflakeCatalogConfigChange(RelationConfigChange):
+    context: Optional[SnowflakeCatalogConfig] = None
+
+    @property
+    def requires_full_refresh(self) -> bool:
+        return True
```
VersusFacit, line 121: This is top of mine for me on incremental models and things for DTs may be different. Is this a safe operation for our users? Is there a way we could cause users to lose data without meaning to or is this designed to gate users behind intentional use of `--full-refresh`?

dbt/include/snowflake/macros/relations/dynamic_table/create.sql
```diff
@@ -1,16 +1,83 @@
...
+--  - sql: str - the code defining the model
+--  Globals:
+--  - config: NodeConfig - contains the attribution required to produce a SnowflakeDynamicTableConfig
+--  Returns:
+--      A valid DDL statement which will result in a new dynamic table.
+-#}
 
     {%- set dynamic_table = relation.from_config(config.model) -%}
 
+    {%- if dynamic_table.catalog.table_format == 'iceberg' -%}
```
VersusFacit, line 18: Why check this manually through the catalog attribute instead of doing a `is_iceberg_format` method call on the relation?

dbt/include/snowflake/macros/relations/dynamic_table/replace.sql
```diff
@@ -1,16 +1,82 @@
...
+--  - sql: str - the code defining the model
+--  Globals:
+--  - config: NodeConfig - contains the attribution required to produce a SnowflakeDynamicTableConfig
+--  Returns:
+--      A valid DDL statement which will result in a new dynamic table.
+-#}
 
     {%- set dynamic_table = relation.from_config(config.model) -%}
 
+    {%- if dynamic_table.catalog.table_format == 'iceberg' -%}
```
VersusFacit, line 18: ditto

tests/functional/relation_tests/dynamic_table_tests/models.py
```diff
@@ -28,6 +28,20 @@
 """
 
 
+DYNAMIC_ICEBERG_TABLE = """
+{{ config(
+    materialized='dynamic_table',
+    snowflake_warehouse='DBT_TESTING',
+    target_lag='2        minutes',
```
VersusFacit, line 35: tab where there should be a space by chance? :)

tests/functional/relation_tests/test_relation_type_change.py
```diff
@@ -1,21 +1,26 @@
...
     relation_type: str
+    table_format: Optional[str] = None
 
     @property
     def name(self) -> str:
-        return f"{self.relation_type}"
+        name = f"{self.relation_type}"
+        if self.table_format:
+            name += f"_{self.table_format}"
+        return name
```
VersusFacit, line 23: It's a trifle in some respect, but I'd aim for a more pythonic solution here.
```suggestion
        if self.table_format:
            return f"{self.relation_type}_{self.table_format}"
        else:
            return f"{self.relation_type}"
```

or even 
```suggestion
    def get_name(self):
        parts = [self.relation_type]
        if self.table_format:
            parts.append(self.table_format)
        return '_'.join(parts)
```

Mutating strings feels unidiomatic so I tend to call it out when I see as a chance to avoid code smell.

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: Oh do keys auto-add themselves on `extra`? I thought that was just a generic `dict`.

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: `relation_results` is a normal `dict`, the values are `agate.Table` instances. It's not save in the sense that we're not validating, or using `get()`. But these `relation_results` should always have a `"catalog"` key. Though, that would be another way to handle the "default" scenario, which actually does feel a bit better. I'll think about swapping it after talking it out.

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: Good catch. This is just an artifact from iteration. I think I originally had `if catalog_results is None or catalog.<something>` and I needed to do the two part check to be able to check the attribution in the second. But then I decided to always return a catalog and just didn't update this. I'll fix it.

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: Users can prevent a full refresh here by specifying `on_configuration_change='continue'` or `on_configuration_change='fail'`. The default is `on_configuration_change='apply'`, which means if we sense a change, we're updating/overwriting. I could see swapping that default in hindsight for the reason you mention, but that was the desired behavior when it was implemented, hence it would be a change in behavior to change this.

dbt/include/snowflake/macros/relations/dynamic_table/create.sql
mikealfare: One of the goals of using `RelationConfig` is to reduce the dependency on the global variables, e.g. `relation` and `config`. If we were able to not rely on any globals, then we could test an individual macro without needing to create any jinja context. So while using `relation.is_iceberg_format` is the right approach on other relation types, I feel like this aligns with the `RelationConfig` approach.

tests/functional/relation_tests/dynamic_table_tests/models.py
mikealfare: This might be outdated since we've confirmed how this works, but the issue I'm implicitly testing is that Snowflake parses this string and stores it in their metadata as `2 minutes`. Honestly, we probably should just say that this is not documented support, hence we are not testing it.

tests/functional/relation_tests/test_relation_type_change.py
mikealfare: I don't think of building strings as "un-pythonic", but if it makes it easier to read what's going on (which I see the argument here), I don't mind changing it. However, I wouldn't have separate `return`s. I would create the name in each branch and then return the name at the end. I'll make that change.

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: What's the earlier syntax suggestion?

dbt/adapters/snowflake/relation_configs/catalog.py
mikealfare: Oh, I just realized what you meant.

tests/functional/relation_tests/dynamic_table_tests/models.py
VersusFacit: We opted against this for now!

dbt/include/snowflake/macros/relations/dynamic_table/create.sql
```diff
@@ -1,16 +1,83 @@
...
+--  - dynamic_table: SnowflakeDynamicTableConfig - contains all of the configuration for the dynamic table
+--  - relation: Union[SnowflakeRelation, str]
+--      - SnowflakeRelation - required for relation.render()
+--      - str - is already the rendered relation name
+--  - sql: str - the code defining the model
+--  Returns:
+--      A valid DDL statement which will result in a new dynamic iceberg table.
+-#}
+
+    create dynamic iceberg table {{ relation }}
```
colin-rogers-dbt, line 73: do we need to call `.render()` if `relation` is a `Relation`?

dbt/include/snowflake/macros/relations/dynamic_table/create.sql
VersusFacit: Makes sense. I like the idea of catalog becoming a first class citizen for dispatching code paths, especially after we talked about this as a team.

tests/functional/relation_tests/test_relation_type_change.py
VersusFacit: Fair enough. Maybe I've just been reading too much new fangled functional Python books of late which I totally accept is an arbitrary choice :)

dbt/include/snowflake/macros/relations/dynamic_table/create.sql
mikealfare: We weren't doing that before this change. I wouldn't want to change that behavior, at least not as part of this PR.

colin-rogers-dbt (approved):

dbt/include/snowflake/macros/utils/optional.sql
```diff
@@ -0,0 +1,14 @@
...
+--
+--  Args:
+--  - name: the name of the DDL option
+--  - value: the value of the DDL option, may be None
+--  - quote_char: the quote character to use (e.g. string), leave blank if unnecessary (e.g. integer or bool)
+--  Returns:
+--      If the value is not None (e.g. provided by the user), return the option setting DDL
+--      If the value is None, return an empty string
+-#}
+{% if value is not none %}{{ name }} = {{ quote_char }}{{ value }}{{ quote_char }}{% endif %}
```
colin-rogers-dbt, line 13: This would be good to move to common or dbt-adapters

mikealfare: Agreed, but I didn't want to tangle with dependent PRs at the moment.

PR #1190, opened 2024-09-24
Add logic to enable iceberg incremental tables.

VersusFacit:
related to #321

### Problem

Add incremental models

### Solution

Leverage existing code abstractions.

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/materializations/incremental.sql
+++ b/dbt/include/snowflake/macros/materializations/incremental.sql
@@ -58,12 +58,20 @@
 
 {% materialization incremental, adapter='snowflake', supported_languages=['sql', 'python'] -%}
 
-  {% set original_query_tag = set_query_tag() %}
 
   {#-- Set vars --#}
   {%- set full_refresh_mode = (should_full_refresh()) -%}
+  {%- set identifier = model['alias'] -%}
   {%- set language = model['language'] -%}
-  {% set target_relation = this %}
+
+  {%- set target_relation = api.Relation.create(
+	identifier=identifier,
+	schema=schema,
+	database=database,
+	type='table',
+	table_format=config.get('table_format', 'default')
+    ) -%}
+
   {% set existing_relation = load_relation(this) %}
 
   {#-- The temp relation will be a view (faster) or temp table, depending on upsert/merge strategy --#}
@@ -90,10 +98,21 @@
     {%- call statement('main', language=language) -%}
       {{ create_table_as(False, target_relation, compiled_code, language) }}
     {%- endcall -%}
+
   {% elif full_refresh_mode %}
+    {% if target_relation.needs_to_drop(existing_relation) %}
+      {{ drop_relation_if_exists(existing_relation) }}
+    {% endif %}
     {%- call statement('main', language=language) -%}
       {{ create_table_as(False, target_relation, compiled_code, language) }}
     {%- endcall -%}
+
+  {% elif target_relation.table_format != existing_relation.table_format %}
+    {% do exceptions.raise_compiler_error(
+        "Unable to convert a " ~ existing_relation.table_format ~ " format table to " ~
+        target_relation.table_format ~ " without specifying --full-refresh."
+      )
+    %}
 
   {% else %}
     {#-- Create the temp relation, either as a view or as a temp table --#}


dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -58,12 +58,20 @@
...
 
-  {% set original_query_tag = set_query_tag() %}
 
   {#-- Set vars --#}
   {%- set full_refresh_mode = (should_full_refresh()) -%}
+  {%- set identifier = model['alias'] -%}
   {%- set language = model['language'] -%}
-  {% set target_relation = this %}
+
+  {%- set target_relation = api.Relation.create(
```
VersusFacit, line 67: this is what we do for plain tables

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -90,11 +98,22 @@
...
+
   {% elif full_refresh_mode %}
+    {% if target_relation.needs_to_drop(existing_relation) %}
+      {{ drop_relation_if_exists(existing_relation) }}
+    {% endif %}
     {%- call statement('main', language=language) -%}
       {{ create_table_as(False, target_relation, compiled_code, language) }}
     {%- endcall -%}
 
+  {% elif target_relation.table_format != existing_relation.table_format %}
```
VersusFacit, line 110: We may ultimately change this behavior (although I'm lobbying for it right now)

dbt/include/snowflake/macros/materializations/table.sql
mikealfare, line 1: I appreciate the consistency in naming.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -58,12 +58,20 @@
 
 {% materialization incremental, adapter='snowflake', supported_languages=['sql', 'python'] -%}
 
-  {% set original_query_tag = set_query_tag() %}
 
   {#-- Set vars --#}
   {%- set full_refresh_mode = (should_full_refresh()) -%}
+  {%- set identifier = model['alias'] -%}
```
mikealfare, line 64: Is `alias` always populated? Or is it an override and we need to pull the name when the user does not provide an alias?

mikealfare (commented):
Does this achieve iceberg support for incremental models? It would be nice if it's this straight forward.

I also noticed that tests have not been added/updated. I would have expected to see some tests with `table_format='iceberg'` in this change.

dbt/include/snowflake/macros/materializations/incremental.sql
VersusFacit: We're doing this for now! We will be looking to smoothly transition tables between formats with the advent of more catalog formalizations in dbt.

dbt/include/snowflake/macros/materializations/incremental.sql
VersusFacit: Great callout! I saw this is how we did it in the table materialization, but I also am prompted to reflect deeper and admit I don't know the subtleties of this model value. Now for sake of time and reducing risk, I am instead opting for `this.name` since that's far safer an option. Then again, the model['alias'] pattern is precisely what we're using in the table materialization, so I assume it's working fine anyways.

tests/functional/relation_tests/test_relation_type_change.py
VersusFacit, line 1: These are Mike's changes I checkedout ahead of time to minimize blast radius of merge conflicts.

tests/functional/relation_tests/test_relation_type_change.py
```diff
@@ -34,31 +40,86 @@ def error_message(self) -> str:
...
 relations = [
     Model(models.VIEW, "view"),
-    Model(models.TABLE, "table"),
-    Model(models.DYNAMIC_TABLE, "dynamic_table"),
+    Model(models.TABLE, "table", "default"),
+    # to be activated upon merge of dynamic table support PR
+    # Model(models.DYNAMIC_TABLE, "dynamic_table", "default"),
+    # Model(models.DYNAMIC_ICEBERG_TABLE, "dynamic_table", "iceberg"),
+    Model(models.ICEBERG_TABLE, "table", "iceberg"),
+    Model(models.ICEBERG_INCREMENTAL_TABLE, "table", "iceberg", incremental=True),
```
VersusFacit, line 48: Technically the way these tests work is by grabbing what Snowflake calls it, which is going to be a table, so it makes sense to use that value here, despite the fact it's from an incremental model

tests/functional/relation_tests/test_relation_type_change.py
```diff
@@ -34,31 +40,86 @@ def error_message(self) -> str:
 
 relations = [
     Model(models.VIEW, "view"),
-    Model(models.TABLE, "table"),
-    Model(models.DYNAMIC_TABLE, "dynamic_table"),
+    Model(models.TABLE, "table", "default"),
+    # to be activated upon merge of dynamic table support PR
```
VersusFacit, line 44: these will be uncommented when we reconcile merge conflicts with Mike's dynamic table PR

tests/functional/relation_tests/test_relation_type_change.py
```diff
@@ -34,31 +40,86 @@ def error_message(self) -> str:
...
+        )
+
+
+class TestRelationTypeChangeBetweenIcebergMaterializations(TestRelationTypeChange):
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {"flags": {"enable_iceberg_materializations": True}}
+
+    @staticmethod
+    def include(scenario) -> bool:
```
mikealfare, line 122: I would roll this scenario into `TestRelationTypeChangeIcebergOn` by removing the `and not scenario.initial.incremental and not scenario.final.incremental` from `include`. The project configuration is the same, so we're really just expending what gets run in the scenario.

The goal of combining all of these into a single test case was to minimize how many times we need to call `dbt run`, since that was the bottleneck for these tests. The reason for two test cases was to support the two different project configuration (the behavior flag). The base test only runs scenarios where iceberg is not a factor. The "iceberg on" test only runs scenarios where iceberg is a factor. Taken together, it's just the 16 (4 starting states X 4 ending states) possible scenarios.

tests/functional/relation_tests/models.py
```diff
@@ -31,3 +31,37 @@
 ) }}
 select * from {{ ref('my_seed') }}
 """
+
+
+DYNAMIC_ICEBERG_TABLE = """
```
mikealfare, line 36: This technically shouldn't exist yet, but would come in when we merge the iceberg DT PR, and/or when we un-comment line 46 below.

tests/functional/iceberg/test_incremental_models.py
```diff
@@ -0,0 +1,126 @@
+import pytest
+
+from pathlib import Path
+
+
+from dbt.tests.util import run_dbt, run_dbt_and_capture, rm_file, write_file
+
+_SEED_INCREMENTAL_STRATEGIES = """
```
mikealfare, line 8: 🚀

tests/functional/iceberg/test_incremental_models.py
```diff
@@ -0,0 +1,126 @@
...
+            "merge.sql": _MODEL_INCREMENTAL_ICEBERG_MERGE,
+            "delete_insert.sql": _MODEL_INCREMENTAL_ICEBERG_DELETE_INSERT,
+        }
+
+    def test_incremental_strategies_build(self, project, setup_class):
+        run_results = run_dbt()
+        assert len(run_results) == 4
+
+    def __check_correct_operations(self, model_name, /, rows_affected, status="SUCCESS"):
+        run_results, stdout = run_dbt_and_capture(
```
mikealfare, line 89: I would update this to `run_results, _`, or even just `run_dbt` instead of `run_dbt_and_capture` since we don't use the logs. Especially since we do use them later on, but from a different call.

tests/functional/iceberg/test_incremental_models.py
```diff
@@ -0,0 +1,126 @@
...
+            ["show", "--inline", f"select * from {{{{ ref('{model_name}') }}}} where world_id = 4"]
+        )
+        assert run_results[0].adapter_response["rows_affected"] == rows_affected
+        assert run_results[0].adapter_response["code"] == status
+
+        if model_name != "append":
+            run_results, stdout = run_dbt_and_capture(
```
mikealfare, line 96: I think you need the first arg to be `"--debug"` when you want to capture the logs. Put another way, the line below `assert "Doughnut" not in stdout` will always be true here because we don't get any logs back.

tests/functional/relation_tests/test_relation_type_change.py
```diff
@@ -34,31 +40,86 @@ def error_message(self) -> str:
...
         run_dbt(["run"])
 
     @pytest.mark.parametrize("scenario", scenarios, ids=[scenario.name for scenario in scenarios])
     def test_replace(self, project, scenario):
-        relation_type = query_relation_type(project, scenario.name)
-        assert relation_type == scenario.final.relation_type, scenario.error_message
+        if self.include(scenario):
+            relation_type = query_relation_type(project, scenario.name)
+            assert relation_type == scenario.final.relation_type, scenario.error_message
+            if relation_type == "dynamic_table":
```
mikealfare, line 87: We're only checking `table_format` on dynamic tables, but we're including standard iceberg tables in the scenarios. I think it's worth creating a "`describe`-like" macro for tables as well to test this, given that one of the goals is to make sure we can create iceberg tables over non-iceberg tables.

The difference in parsing can be handled directly in a new `describe_table` helper in `utils`, akin to `describe_dynamic_table` (i.e. I don't think there's a need to create a whole RelationConfig for tables as there is for DTs just to test this).

tests/functional/iceberg/test_incremental_models.py
VersusFacit: That's just a mis-copy there :) Good catch

tests/functional/relation_tests/models.py
VersusFacit: I just copied the whole file via your commit to avoid merge conflicts 🫡

tests/functional/relation_tests/test_relation_type_change.py
VersusFacit: I think we should do this, but this is covered implicitly by Iceberg table tests where a swap is done.

tests/functional/iceberg/test_incremental_models.py
VersusFacit: We talked about this -- this is fine :D

mikealfare (approved):

VersusFacit:
I verified that the new configuration of the replacement type tests is replacing an iceberg table with an incremental iceberg table and vice versa, even after we changed the test case logic.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -58,12 +58,21 @@
 
 {% materialization incremental, adapter='snowflake', supported_languages=['sql', 'python'] -%}
 
-  {% set original_query_tag = set_query_tag() %}
```
francois-blanchard, line 61: Hi @VersusFacit, I noticed that the query tags for incremental models are not applied anymore.
Why did you remove this line?

PR #1192, opened 2024-09-26
[Microbatch] Optimizations: use view for temp relation + remove `using` clause during delete statement

### LINKED ISSUE #1228: [Bug] Cartesian Join based deletion is causing performance problems when it hits a certain scale for microbatch models
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core

### Current Behavior

 Cartesian Join based deletion is causing data spilling to disk which heavily bogs down performance

The delete statement looks like:
```
delete from analytics_dev.dbt_aescay.my_model DBT_INTERNAL_TARGET
    using analytics_dev.dbt_aescay.my_model__dbt_tmp DBT_TMP_TARGET
    where (
    DBT_INTERNAL_TARGET.event_at >= TIMESTAMP '2024-10-14 00:00:00+00:00'
    and DBT_INTERNAL_TARGET.event_at < TIMESTAMP '2024-10-15 00:00:00+00:00'
    
    );
```

But we are not doing anything with `my_model__dbt_tmp` in the `where` clause. 

We can simplify this logic and improve the performance, by instead doing:
```
delete from <existing> where <date range>;
insert into <existing> from <new data for same date range>;
``` 

One advantage of microbatch is that we know in advance the exact boundaries of every batch (time range, cf. "static" `insert_overwrite`).

In a world where we support "microbatch merge" models (= update batches by upserting on `unique_key`, rather than full batch replacement), then we would want to join (using) based on `unique_key` match, like so:
```
delete from analytics_dev.dbt_aescay.my_model DBT_INTERNAL_TARGET
    using analytics_dev.dbt_aescay.my_model__dbt_tmp DBT_TMP_TARGET
    where DBT_INTERNAL_TARGET.event_id = DBT_TMP_TARGET.event_id
    and (
    DBT_INTERNAL_TARGET.event_at >= TIMESTAMP '2024-10-14 00:00:00+00:00'
    and DBT_INTERNAL_TARGET.event_at < TIMESTAMP '2024-10-15 00:00:00+00:00'
    
    );
```

But this shouldn't be the default assumption.

### Expected Behavior

We should delete [this](https://github.com/dbt-labs/dbt-snowflake/blob/7dea1458ba4a63bafa2f4188768f1059b09dabdd/dbt/include/snowflake/macros/materializations/merge.sql#L69) line. 

### Steps To Reproduce

See [here](https://www.notion.so/dbtlabs/Testing-out-the-Microbatch-Incrementals-121bb38ebda78066bdeff09870d39d31?pvs=4#121bb38ebda780e48dc8d1713508abf5).

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

snowflake

### Additional Context

_No response_

MichelleArk:
resolves #1228
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/# N/A

### Problem

Microbatch performance is underwhelming for dbt-snowflake! A couple underlying reasons:
1) Microbatch largely inherits from delete+insert strategy in snowflake, which has been optimized to use a temp view instead of table. However, microbatch is still using a temp table! More details: https://docs.getdbt.com/reference/resource-configs/snowflake-configs#temporary-tables
  * Note that it is always safe to do so for microbatch since there is no `unique_key` necessary
2) The `using` clause is unnecessary in the microbatch `delete` statement, because microbatch does not require a merge key and the `where` clause is sufficient for a safe deletion.

### Solution

1) Apply temp view optimization to microbatch. Was just a matter of extending a couple if statements that only checked for insert+delete previously
2) Remove extraneous `using` clause

### Checklist


dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -49,7 +49,7 @@
     {{ return("view") }}
   {% elif strategy in ("default", "merge", "append") %}
     {{ return("view") }}
-  {% elif strategy == "delete+insert" and unique_key is none %}
+  {% elif strategy in ["delete+insert", "microbatch"] and unique_key is none %}
```
QMalcolm, line 52: 🚀 This is a greeeeaaaat catch! Nice!

QMalcolm (approved):
Looks good to me! :shipit:

colin-rogers-dbt (approved):

QMalcolm (approved):

mikealfare (approved):

PR #1197, opened 2024-10-02
Update database role tests to avoid colliding with parallel test runs

mikealfare:
### Problem

The affected database role test uses a static name for the role name. It's possible for two parallel runs to collide and attempt to drop/create the same role at the same time, causing the test to fail.

The affected database role is granted future permissions on all tables in the database, which affects other grants tests.

### Solution

- Add a semi-unique suffix to the role name (1 in 1MM chance of collision now)
- Limit the future grants to just the test schema

### Checklist

### DIFF
--- a/tests/functional/auth_tests/test_database_role.py
+++ b/tests/functional/auth_tests/test_database_role.py
@@ -55,7 +55,7 @@ class TestDatabaseRole:
         sql = f"""
         GRANT
             ALL PRIVILEGES ON FUTURE TABLES
-            IN DATABASE {project.database}
+            IN SCHEMA {project.test_schema}
             TO DATABASE ROLE {role}
         """
         project.run_sql(sql)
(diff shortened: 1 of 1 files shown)


tests/functional/auth_tests/test_database_role.py
```diff
@@ -48,12 +49,13 @@ def setup(self, project):
         The existence of this database role triggered the bug as dbt-snowflake attempts
         to revoke it if the user also provides a grants config.
         """
-        role = "BLOCKING_DB_ROLE"
+        unique_suffix = randrange(0, 1_000_000)
+        role = f"BLOCKING_DB_ROLE_{unique_suffix}"
```
colin-rogers-dbt, line 53: should we use the invocation_id that's tied to the test as a suffix?

mikealfare: We don't have access to the invocation_id, but I can use the test schema prefix.

colin-rogers-dbt (approved):

PR #1201, opened 2024-10-08
Fix Dynamic Iceberg Table Required DDL Params

### LINKED ISSUE #1200: [Bug] dynamic iceberg tables don't build without base_location_subpath option

### Current Behavior

`base_location_subpath` should be an optional parameter that the user provides as an additional scoping mechanism. Instead, it is currently responsible for base generation creation.

image

### Expected Behavior

Dynamic iceberg tables should render ddl using the `_dbt/schema/relation name` base location when no `base_location_subpath` is provided.

### Steps To Reproduce

1. in an iceberg-enabled snowflake, write a dynamic iceberg table code artifact with no `base_location_subpath` param
2. `dbt run`

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-snowflake:
```

### Additional Context

_No response_

VersusFacit:
resolves #1200 

### Problem

Base location isn't generating a default value.

### Solution

Render the base location explicitly.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241008-122635.yaml
+++ b/.changes/unreleased/Fixes-20241008-122635.yaml
@@ -1,5 +1,5 @@
 kind: Fixes
-body: Dynamic Iceberg table DDL catalog parameter fixes.
+body: Dynamic Iceberg table base_location_subpath generation fix.
 time: 2024-10-08T12:26:35.521308-07:00
 custom:
     Author: versusfacit
--- a/.changes/unreleased/Fixes-20241008-122635.yaml
+++ b/.changes/unreleased/Fixes-20241008-122635.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Dynamic Iceberg table DDL catalog parameter fixes.
+time: 2024-10-08T12:26:35.521308-07:00
+custom:
+    Author: versusfacit
+    Issue: "1200"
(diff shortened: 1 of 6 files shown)


dbt/include/snowflake/macros/relations/dynamic_table/create.sql
```diff
@@ -73,9 +73,9 @@
     create dynamic iceberg table {{ relation }}
         target_lag = '{{ dynamic_table.target_lag }}'
         warehouse = {{ dynamic_table.snowflake_warehouse }}
-        {{ optional('external_volume', dynamic_table.catalog.external_volume) }}
-        {{ optional('catalog', dynamic_table.catalog.name) }}
-        base_location = {{ dynamic_table.catalog.base_location }}
+        external_volume = '{{ dynamic_table.catalog.external_volume }}'
```
VersusFacit, line 76: these aren't optional so I'm removing that

.changes/unreleased/Fixes-20241008-122635.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Dynamic Iceberg table DDL catalog parameter fixes.
```
mikealfare, line 2: I would consider making this changelog entry specific to `base_location_subpath`.

mikealfare (approved):
One, non-blocking consideration.

PR #1209, opened 2024-10-16
Add snowflake telemetry.

VersusFacit:
Related to https://github.com/dbt-labs/dbt-adapters/issues/301

### Problem

We need a snowflake override to get the data for snowflake.

### Solution

Override the base class method.

### Checklist

### DIFF
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -1,9 +1,12 @@
 from dataclasses import dataclass
+from importlib import import_module
 from typing import Mapping, Any, Optional, List, Union, Dict, FrozenSet, Tuple, TYPE_CHECKING
 
 from dbt.adapters.base.impl import AdapterConfig, ConstraintSupport
 from dbt.adapters.base.meta import available
+from dbt.adapters.base.relation import AdapterTrackingRelationInfo
 from dbt.adapters.capability import CapabilityDict, CapabilitySupport, Support, Capability
+from dbt.adapters.contracts.relation import RelationConfig
 from dbt.adapters.sql import SQLAdapter
 from dbt.adapters.sql.impl import (
     LIST_SCHEMAS_MACRO_NAME,
@@ -25,6 +28,7 @@ from dbt.adapters.snowflake.relation_configs import (
     SnowflakeRelationType,
     TableFormat,
 )
+
 from dbt.adapters.snowflake import SnowflakeColumn
 from dbt.adapters.snowflake import SnowflakeConnectionManager
 from dbt.adapters.snowflake import SnowflakeRelation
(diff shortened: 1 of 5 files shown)


VersusFacit:
image

The core function is using the Snowflake override successfully :D 

My local env for testing.

image

dbt/adapters/snowflake/impl.py
```diff
@@ -419,3 +423,21 @@ def valid_incremental_strategies(self):
     def debug_query(self):
         """Override for DebugTask method"""
         self.execute("select 1 as id")
+
+    def get_adapter_run_info(cls, config: RelationConfig) -> AdapterTrackingRelationInfo:
```
mikealfare, line 427: See my comment in the associated PR on `dbt-adapters`. I think this could be simplified to:

```python
@classmethod
def _get_adapter_run_info(cls, config) -> Dict[str, Any]:
    if (
        config
        and hasattr(config, "_extra")
        and (relation_format := config._extra.get("table_format"))
    ):
        return relation_format
    return None
```
Two more things, the `@classmethod` decorator is missing, and I didn't realize you could use the walrus operator in a compound `if` statement like that. I've found a few scenarios where I wanted to do that (like this scenario, when checking if the attribute exists first) and I either couldn't get it to work, or just assumed it didn't. That's nice to know!

VersusFacit: I didn't think I needed the decorator here. But I'll add it for continuity.

colin-rogers-dbt (approved):

dbt/adapters/snowflake/impl.py
```diff
@@ -419,3 +421,18 @@ def valid_incremental_strategies(self):
     def debug_query(self):
         """Override for DebugTask method"""
         self.execute("select 1 as id")
+
+    @classmethod
+    def _get_adapter_specific_run_info(cls, config: RelationConfig) -> Dict[str, Any]:
```
mikealfare, line 426: Do we want `None`, `""`, or to omit `table_format` entirely when it doesn't exist? Does this output align with the jsonschema? (it may, just validating)

mikealfare (commented):
Besides the one question, is it worth adding a test to this since we are tracking an additional `dbt-snowflake`-specific attribute?

dbt/adapters/snowflake/impl.py
VersusFacit: It does align with the Json schema now that I've added the null case to the upstream. Thanks for pointing this out

VersusFacit:
I suppose it's worth adding a unit test (done) because this gives us extra security with base adapter / core calls that could be tripped up through many layers of indirection.

PR #1210, opened 2024-10-16
adding transient dynamic table capabilities

jhsb25:
resolves #

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20241017-103843.yaml
+++ b/.changes/unreleased/Features-20241017-103843.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Adding capability to create transient dynamic tables.
+time: 2024-10-17T10:38:43.820283+01:00
+custom:
+    Author: jhsb25
+    Issue: "1089"
--- a/dbt/adapters/snowflake/impl.py
+++ b/dbt/adapters/snowflake/impl.py
@@ -92,7 +92,18 @@ class SnowflakeAdapter(SQLAdapter):
                     "benefits only those actively using it, we've made this behavior opt-in to "
                     "prevent unnecessary latency for other users."
                 ),
-            }
+            },
+            {
+                "name": "default_dynamic_tables_to_transient",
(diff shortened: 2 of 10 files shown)


jhsb25:
@cla-bot check

bpruss:
I'm new here, is this an appropriate place to inquire about this PR and the issue it is trying to fix. 
I had been following https://github.com/dbt-labs/dbt-adapters/issues/716 and was really optimisitic it had made it into 1.9.  
I saw multiple approval requests from jshb25. I was so optimistic that I added `transient=true,` to my dynamic table macro only to find that it didn't work.  Did some searching and ended up here. Thanks for any info that can be provided.

hawkinsrt:
I'm also curious about the status of this PR.  We use dynamic tables for our analytic data marts but do not need the ability to time travel.  It would be nice to avoid that cost and have them created as dynamic transient tables.

anth-zen:
Sorry if this is inappropriate due to its sketchiness, but I wanted to share my hack/workaround in case it helps anyone else save a lot of money in Snowflake.

Disclaimer: The following hack is a generally horrible thing to do unless you're desperate, and it may cause other problems.

I was able to directly edit my local virtual environment to hard-wire the word "transient" into the dbt dynamic table macros. This forces dbt to always materialize dynamic tables as transient in your local dbt runs. This surely won't work in a default CI/CD pipeline, but for local development it is saving us a lot of Snowflake Fail-safe costs.

(I'm using dbt 1.10.3 and snowflake 1.9.4, so these references may be incorrect for other versions.)

**In folder**: `.venv\Lib\site-packages\dbt\include\snowflake\macros\relations\dynamic_table`
* **create.sql**
  * Change line 34 to say: `create transient dynamic table`
* **replace.sql**
  * Change line 47 to say: `create or replace transient dynamic table`

Note: These changes will certainly get reverted if you re-install or upgrade dbt locally, so you'll need to re-apply them.

PR #1212, opened 2024-10-17
Add query tag back to incremental models

mikealfare:
### Problem

We inadvertently removed query tags on incremental models when implementing Iceberg support.

### Solution

- Add query tags back to incremental models
- Fix the query tag test that should have caught this, but didn't

I'm skipping the changelog on this since this is a pre-regression that we caught prior to final release.

### Checklist

### DIFF
--- a/tests/functional/query_tag/test_query_tags.py
+++ b/tests/functional/query_tag/test_query_tags.py
@@ -1,6 +1,7 @@
 import pytest
 from dbt.tests.util import run_dbt
 
+
 snapshots__snapshot_query_tag_sql = """
 {% snapshot snapshot_query_tag %}
     {{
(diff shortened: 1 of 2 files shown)


tests/functional/query_tag/test_query_tags.py
```diff
@@ -95,20 +93,14 @@ def seeds(self):
...
-            "models": {
-                "tests": {"query_tag": prefix, "post-hook": "{{ check_tag() }}"},
-            },
-            "seeds": {
-                "tests": {"query_tag": prefix, "post-hook": "{{ check_tag() }}"},
-            },
-            "snapshots": {
-                "tests": {"query_tag": prefix, "post-hook": "{{ check_tag() }}"},
-            },
+            "models": {"query_tag": prefix, "post-hook": "{{ check_query_tag() }}"},
```
mikealfare, line 96: If you look at the successful runs, they warn that `models.tests`, `seeds.tests`, and `snapshots.tests` were unused configuration paths. In other words, the post-hook never ran. We also know that since the post-hook references a macro that does not exist; it should have referenced `check_query_tag` instead of `check_tag`.

mikealfare:
CI is failing due to the pending update for py38 retirement and the related change in branch protection rules. This should be resolved once the py38 retirement PR is merged, which contain the expected workflows.

colin-rogers-dbt (approved):

PR #1214, opened 2024-10-18
fix: [microbatch] delete redundant using clause

### LINKED ISSUE #1198: [Feature] Incremental strategies for delete+insert and microbatch cause unnecessary cross joins

### Describe the feature

Currently, the delete+insert incremental strategy requires a "unique_key", but there's nothing actually enforcing this key be unique. I've done this before, but the issue is that it creates a cartesian join between all the records with matching (non-unique) unique_key values because of the USING. The new microbatch strategy does something similar, where if you don't have an additional predicate, the USING <source> bit is unnecessary. In both cases, the join explosion can take up a significant amount of resources.

For the delete+insert strategy, I would propose possibly changing the name the "unique_key" config to "incremental_key", and modifying the query to use a correlated EXISTS subquery.

```sql
delete from {{ target }}
where exists (
    select 1
    from {{ source }}
    where
    {% if incremental_key is sequence and incremental_key is not string %}
        {% for key in incremental_key %}
            {{ source }}.{{ key }} = {{ target }}.{{ key }}
            {{ "and " if not loop.last}}
        {% endfor %}
    {% else %}
        {{ source }}.{{ incremental_key }} = {{ target}}.{{ incremental_key }}
    {% endif %}
)
{% if incremental_predicates %}
    {% for predicate in incremental_predicates %}
        {{ predicate }} {{ "and" if not loop.last }}
    {% endfor %}
{% endif %};
```

The `using {{ source }}` line in the microbatch strategy should simply be deleted.

### Who will this benefit?

This would benefit every who uses these incremental strategies by significantly lowering warehouse load (and therefore costs) for these types of queries.

I think the code above should work. It may take me a while before I have time to do all the requisite testing before creating a PR.

### Anything else?

I ran one query like `DELETE FROM target USING source` with no additional predicates where the target table had 10 million records and the source table a billion. You can see how relatively expensive that cartesian join is.

Most Expensive Nodes
CartesianJoin 47.3%
Delete 27.0%
TableScan 1.3%

ugmuka:
resolves #1198

### Problem
The microbatch feature generates a DELETE SQL statement that includes a redundant USING clause. The current SQL looks like this:

```
delete from <target_model_name> DBT_INTERNAL_TARGET
    using <tmp_table>
    where (
    DBT_INTERNAL_TARGET.event_date >= TIMESTAMP '2024-10-13 00:00:00+00:00'
    and DBT_INTERNAL_TARGET.event_date < TIMESTAMP '2024-10-14 00:00:00+00:00'
    
    );
```

This results in a Cartesian join, which is unnecessary. To resolve this issue, the USING <tmp_table> clause should be removed.

### Solution
Remove the USING <tmp_table> clause from merge.sql.

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/materializations/merge.sql
+++ b/dbt/include/snowflake/macros/materializations/merge.sql
@@ -66,7 +66,6 @@
     {% do arg_dict.update({'incremental_predicates': incremental_predicates}) %}
 
     delete from {{ target }} DBT_INTERNAL_TARGET
-    using {{ source }}
     where (
     {% for predicate in incremental_predicates %}
         {%- if not loop.first %}and {% endif -%} {{ predicate }}


ugmuka:
I have signed CLA.

colin-rogers-dbt:
@cla-bot check

MichelleArk:
Closing in favor of an equivalent fix that went out in: https://github.com/dbt-labs/dbt-snowflake/pull/1192/files#diff-03ee8d11934e66a23fa4ab1a018378a9533816c757ed0c709fad89ada1d12d79L69

Issue #1234, opened 2024-11-07
[Bug] `dbt-snowflake` raises an error when caching less than the maximum number of objects in a schema


mikealfare:
### Is this a new bug in dbt-snowflake?

- [X] I believe this is a new bug in dbt-snowflake
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

I'm seeing this error when running `dbt-snowflake` on a schema with 12,500 objects:
```shell
16:39:51  Encountered an error:
Compilation Error in macro list_relations_without_caching (macros/adapters/metadata.sql)
  dbt will list a maximum of 100000 objects in schema <DATABASE>.<SCHEMA>.
             Your schema exceeds this limit. Please contact support@getdbt.com for troubleshooting tips,
             or review and reduce the number of objects contained.
  
  > in macro snowflake__get_paginated_relations_array (macros/adapters.sql)
  > called by macro snowflake__list_relations_without_caching (macros/adapters.sql)
  > called by macro list_relations_without_caching (macros/adapters/metadata.sql)
  > called by macro list_relations_without_caching (macros/adapters/metadata.sql)
```

I can confirm that there are less than 100,000 objects in this schema.
```
-- use dbt service account role
show objects in <DATABASE>.<SCHEMA>;
result: 12,500
```

### Expected Behavior

I should be able to run `dbt-snowflake` on a schema with 12,500 objects and the cache should reflect all 12,500 objects with no error.

### Steps To Reproduce

1. Create a project with 12,500 objects in a single schema (or more than 10K really)
2. Run `dbt run` on the full project

### Environment

```markdown
- OS: dbt Cloud
- Python: 3.9
- dbt-core: versionless
- dbt-snowflake: versionless
```

### Additional Context

Initial analysis suggests this is the result of the [2024_07](https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_07_bundle) Snowflake bundle. In particular, it looks like part of this [change](https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_07/bcr-1735). We use the object name as a watermark when paginating. This is no longer deterministic, meaning that we could (and very likely do) get duplicates of relations and miss relations because we're picking a random starting point in a randomized list. In the past, this return recordset was ordered, making the pagination deterministic. We need to figure out a new watermark method.

PR #1235, opened 2024-11-07
Make `list_relations_without_caching` pagination configurable

mikealfare:
resolves #1234

### Problem

Some users have over 100K objects in a single schema. We need to paginate results from `show objects` since that call is limited to 10K objects by Snowflake. We hard coded this result to 10 pages. However this doesn't allow users to have more than 100K objects in their schema. In this scenario, users are hit with an error that blocks dbt from running anything.

### Solution

Allow users to configure the pagination using project flags. They can configure both the number of results per page as well as the number of pages. This defaults to the current setting of 10000 results per page and 10 pages if the flags are not passed in. Example:

```yml
# dbt-project.yml
flags:
  list_relations_per_page: 10000
  list_relations_page_limit: 100
```

### Checklist

### DIFF
--- a/dbt/include/snowflake/macros/adapters.sql
+++ b/dbt/include/snowflake/macros/adapters.sql
@@ -111,9 +111,10 @@
 
       {%- if loop.index == max_iter -%}
         {%- set msg -%}
-           dbt will list a maximum of {{ max_total_results }} objects in schema {{ schema_relation }}.
-           Your schema exceeds this limit. Please contact support@getdbt.com for troubleshooting tips,
-           or review and reduce the number of objects contained.
+            dbt is currently configured to list a maximum of {{ max_total_results }} objects per schema.
+            {{ schema_relation }} exceeds this limit. If this is expected, you may configure this limit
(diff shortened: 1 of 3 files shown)


colin-rogers-dbt (approved):

dbt/include/snowflake/macros/adapters.sql
```diff
@@ -134,7 +135,10 @@
 
 {% endmacro %}
 
-{% macro snowflake__list_relations_without_caching(schema_relation, max_iter=10, max_results_per_iter=10000) %}
+{% macro snowflake__list_relations_without_caching(schema_relation) %}
+
+  {%- set max_results_per_iter = adapter.config.flags.get('list_relations_per_page', 10000) -%}
```
VersusFacit, line 140: Did we consult DX or PM folks on the names here? I'm just not sure if it's the most ergonomic.

VersusFacit (commented):
More curious here than anything else: is there any real cost on this test from building all of these views? I imagine it's quite like since we're not actually running anything per se.

mikealfare:
I think this got accidentally closed with the docs PR.

PR #1238, opened 2024-11-08
Update unit test mocks to handle the new `query_id` field

mikealfare:
### Problem

We added `query_id` as a new field on the `AdapterResponse` event. This needs to be a string to pass validation.

### Solution

Mock the `sfqid` method to return a string literal instead of the mocked object it gets via `MagicMock`.

### Checklist


mikealfare:
Kick the CI

PR #1259, opened 2024-11-27
support MicrobatchConcurrency capability

MichelleArk:
resolves https://github.com/dbt-labs/dbt-snowflake/issues/1260

### Problem

dbt-snowflake does not yet support running the microbatch incremental strategy in concurrent threads.

### Solution

* Ensure temp tables are written to unique locations when `batch.id` is available (in microbatch model context)
* Enable MicrobatchConcurrency capability

### Checklist


dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -80,7 +80,9 @@
   {%- set unique_key = config.get('unique_key') -%}
   {% set incremental_strategy = config.get('incremental_strategy') or 'default' %}
   {% set tmp_relation_type = dbt_snowflake_get_tmp_relation_type(incremental_strategy, unique_key, language) %}
-  {% set tmp_relation = make_temp_relation(this).incorporate(type=tmp_relation_type) %}
+
+  {% set tmp_relation_suffix = '__dbt_tmp' if not model.batch else '__dbt_tmp_' ~ model.batch.id %}
+  {% set tmp_relation = make_temp_relation(this, suffix=tmp_relation_suffix).incorporate(type=tmp_relation_type) %}
```
QMalcolm, line 85: This is perfect and will continue to work after the planned dbt-adapters change!

QMalcolm (approved):
:shipit:

MichelleArk:
📉 🎩 I've done a bunch of benchmarking around this enablement!

### Setup 
* Microbatch input model with several years of data, although 30 days is all that was strictly necessary here. 
* Microbatch model configured to have a 'day' batch size, producing a variable sized batch (1, 1k, 10k, 50k records)
* Running `dbt run --select microbatch --full-refresh --event-time-start 1998-07-01 --event-time-end 1998-08-02` with concurrency disabled, enabled and with 4 threads, enabled and with 8 threads

### Results
#### batch size = 1 row
```
| # concurrent threads          | overall time (s) | sum of batch time (s) | avg time per batch (s) | % of time spent in concurrency-related overhead |
|-------------------------------|------------------|-----------------------|------------------------|-------------------------------------------------|
| 1 (concurrent_batches: False) | 53.96            | 50.98                 | 1.6                   | N/A                                             |
| 4                             | 28.35            | 87.17                 | 2.7                   | 40%                                             |
| 8                             | 22.21            | 120.96                | 3.8s                  | 57%    
```

#### batch size = 1k rows
^ no significant difference to 1 row result

#### batch size = 10k rows
```
| # concurrent threads          | overall time (s) | sum of batch time (s) | avg time per batch (s) | % of time spent in concurrency-related overhead |
|-------------------------------|------------------|-----------------------|------------------------|-------------------------------------------------|
| 1 (concurrent_batches: False) | 75.32            | 72.31                 | 2.25                   | N/A                                             |
| 4                             | 31.02            | 99.94                 | 3.1                    | 27%                                             |
| 8                             | 28.4             | 140.78                | 4.4                    | 48%     
```


#### batch size = 50k rows
```
| # concurrent threads          | overall time (s) | sum of batch time (s) | avg time per batch (s) | % of time spent in concurrency-related overhead |
|-------------------------------|------------------|-----------------------|------------------------|-------------------------------------------------|
| 1 (concurrent_batches: False) | 98.79            | 95.99                | 3                   | N/A                                             |
| 4                             | 39.63           | 131.33                 | 4                    | 25%                                             |
| 8                             | 30.80             | 172.94                | 5.4                   | 44%     
```

Validation: 
* For each scenario, I confirmed that the resulting dataset when running in concurrent mode is equivalent to the serial dataset produced. I've also manually confirmed the temp table suffixing is working as expected (e.g. `create or replace  temporary view analytics.dbt_marky.microbatch__dbt_tmp_19980801`).

### Conclusions

It's clear that there is some amount of overhead associated with enabling concurrency (longer avg runs for individual batches), with the benefit of significantly reduced. It appears that this overhead is non-linear however, and that it should decline relative to the overall runtime the larger the batch to be merged is.

I believe we should proceed with enabling concurrency for dbt-snowflake, and document that the benefit of running models concurrently is faster overall backfilling capabilities, with the side effect of longer-running batches, as some amount of overhead is incurred by the platform to merge into the main dataset safely + concurrently.

dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -80,7 +80,9 @@
   {%- set unique_key = config.get('unique_key') -%}
   {% set incremental_strategy = config.get('incremental_strategy') or 'default' %}
   {% set tmp_relation_type = dbt_snowflake_get_tmp_relation_type(incremental_strategy, unique_key, language) %}
-  {% set tmp_relation = make_temp_relation(this).incorporate(type=tmp_relation_type) %}
+
+  {% set tmp_relation_suffix = '__dbt_tmp' if not model.batch else '__dbt_tmp_' ~ model.batch.id %}
```
MichelleArk, line 84: In order to leverage the work done in https://github.com/dbt-labs/dbt-adapters/pull/361, we'll need to cut a new dbt-adapters release + bump the dbt-adapters minimum dependency in dbt-snowflake.

I'm happy to do that refactor as a follow-on item to decouple those releases. Up to you @mikealfare

mikealfare: Do we need this if we bump the dep on `dbt-adapters`? I think this would happen automatically with that bump. I'm sure I'm missing something.

MichelleArk: We don't, which is what I'm getting at. I can proceed with this change + remove it later once dbt-adapters have been released, or wait for the release and remove the `tmp_relation_suffix` construction here.

mikealfare: We released the patched version of `dbt-adapters`, so let's remove this, per internal conversation.

mikealfare (approved):

